PR #3, opened 2019-01-08
Add array content serialization for website

wachterjohannes:


Tests/Unit/Model/Content/ContentViewTest.php
```diff
@@ -79,9 +79,26 @@ public function testGetData(): void
...
+            'en',
+            'type-1',
+            ['title' => 'Sulu', 'url' => '/test', 'article' => '<p>Sulu is awesome</p>']
+        );
+
+        $result = $contentView->withResource('product', '123-123-123', 'en');
+        $this->assertNotSame($contentView, $result);
+        $this->assertSame('product', $result->getResourceKey());
+        $this->assertSame('123-123-123', $result->getResourceId());
+        $this->assertSame('en', $result->getLocale());
```
nnatter: we should probably check if the other properties of the two objects match?

Tests/Functional/EventSubscriber/ContentViewSerializerSubscriberTest.php
```diff
@@ -0,0 +1,130 @@
+<?php
```
nnatter: ContentView*Serialization*Subscriber

Tests/Application/config/templates/pages/default.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" ?>
```
nnatter: can we use `test_resources/default.xml` here?

PR #19, opened 2019-02-17
Split up circleci configuration into multiple jobs and added code-coverage check

nnatter:


nnatter:
looks like circleci hides the artifacts tab for anonymous users. artifacts are still accessible by appending `#artifacts` to the job-url. (for example `https://circleci.com/gh/sulu/SuluContentBundle/250#artifacts`)

PR #32, opened 2019-10-24
Add Dimension Model with Doctrine Entity

alexander-schranz:
The dimension Entity will make it possible that a content can be created in different dimensions which will be merged into one.


DependencyInjection/SuluContentExtension.php
```diff
@@ -13,17 +13,47 @@
...
+    {
+        if ($container->hasExtension('doctrine')) {
+            $container->prependExtensionConfig(
+                'doctrine',
+                [
+                    'orm' => [
+                        'mappings' => [
+                            'SuluContentBundleDimension' => [
+                                'type' => 'xml',
+                                'prefix' => 'Sulu\Bundle\ContentBundle\Dimension\Domain\Model',
```
nnatter: would propose to use `Domain\Entity` as the doctrine bundle would use an `Entity` directory by default
https://symfony.com/doc/current/reference/configuration/doctrine.html#custom-mapping-entities-in-a-bundle

Dimension/Domain/Model/Dimension.php
```diff
@@ -0,0 +1,73 @@
...
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\ContentBundle\Dimension\Domain\Model;
```
nnatter: this would mean that we have a `Dimension` context, right? what other contexts would you use?

Dimension/Domain/Model/Dimension.php
```diff
@@ -0,0 +1,73 @@
...
+    {
+        return $this->id;
+    }
+
+    public function getLocale(): ?string
+    {
+        return $this->locale;
+    }
+
+    public function setLocale(?string $locale): DimensionInterface
```
nnatter: not sure if this class should provide setters. sounds kind of dangerous to me to change these things for existing dimensions

Tests/coverage-checker.php
```diff
@@ -143,9 +143,11 @@ function printCodeCoverageReport(Directory $pathReport): void
 
 function getReportForPath(Directory $rootReport, string $path): ?Directory
```
nnatter: we should definitely move this classes into a package sometime 🙈

DependencyInjection/SuluContentExtension.php
alexander-schranz: As discussed this namespace should not todo with doctrine so we use Model here.

Dimension/Domain/Model/Dimension.php
alexander-schranz: @nnatter only Dimension and Content. Not sure if we change this when implementing Content.

Dimension/Domain/Model/Dimension.php
alexander-schranz: Correct this should not have setters

Dimension/Domain/Model/Dimension.php
nnatter: would probably use a single context only. things get cumbersome with small contexts really fast 🙂 but we can change that when implementing the Content model

Tests/coverage-checker.php
alexander-schranz: https://github.com/handcraftedinthealps/ namespace?

alexander-schranz:
@nnatter did refractor by renaming published to publishState as I think in future we will maybe have a `review` state or something similiar.

Dimension/Domain/Model/Dimension.php
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * @var string|null
+     */
+    private $locale;
+
+    /**
+     * @var string
+     */
+    private $publishState = DimensionInterface::PUBLISH_STATE_DRAFT;
```
nnatter: `$workflowStage` 😇

Tests/coverage-checker.php
nnatter: 👍

Dimension/Domain/Model/Dimension.php
alexander-schranz: Done

PR #33, opened 2019-10-25
Implement content entities

alexander-schranz:
This will add the AbstractContent and AbstractContentDimension entities which will implement the logic for entities which have Template, Seo and Excerpt.


Content/Domain/Model/AbstractContent.php
```diff
@@ -0,0 +1,55 @@
...
+    public function addDimension(ContentDimensionInterface $contentDimension): void
+    {
+        $this->initializeDimensions();
+
+        $this->dimensions->add($contentDimension);
+    }
+
+    public function removeDimension(ContentDimensionInterface $contentDimension): void
+    {
+        $this->initializeDimensions();
```
nnatter: is there a reason for not doing this in the constructor?

Content/Infrastructure/Doctrine/MetadataLoader.php
```diff
@@ -0,0 +1,176 @@
...
+    private $tagClass;
+
+    /**
+     * @var string
+     */
+    private $categoryClass;
+
+    public function __construct(string $tagClass, string $categoryClass)
+    {
+        $this->tagClass = $tagClass;
```
nnatter: why do you need to inject the `tagClass` but not the `mediaClass`?

Content/Infrastructure/Doctrine/MetadataLoader.php
```diff
@@ -0,0 +1,176 @@
...
+                    'references' => [
+                        'entity' => MediaInterface::class,
+                        'field' => 'id',
+                        'onDelete' => 'CASCADE',
+                        'onUpdate' => 'CASCADE',
+                    ],
+                ],
+            ]);
+
+            $this->addManyToMany($event, $metadata, 'excerptTags', $this->tagClass, 'tag_id');
```
nnatter: i think we need an intermediate entity here to make this sortable in the future?

Content/Infrastructure/Doctrine/MetadataLoader.php
```diff
@@ -0,0 +1,176 @@
...
+
+        $metadata->mapField(array_merge([
+            'fieldName' => $name,
+            'columnName' => $name,
+            'type' => $type,
+            'nullable' => $nullable,
+        ], $mapping));
+    }
+
+    private function getRelationTableName(ClassMetadataInfo $metadata, string $relationName): string
```
nnatter: 😂

Tests/Unit/Content/Infrastructure/Doctrine/MetadataLoaderTest.php
```diff
@@ -0,0 +1,150 @@
...
+use Prophecy\Argument;
+use Sulu\Bundle\CategoryBundle\Entity\Category;
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\ExcerptInterface;
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\SeoInterface;
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\TemplateInterface;
+use Sulu\Bundle\ContentBundle\Content\Infrastructure\Doctrine\MetadataLoader;
+use Sulu\Bundle\ContentBundle\Tests\Application\ExampleTestBundle\Entity\ExampleDimension;
+use Sulu\Bundle\TagBundle\Entity\Tag;
+
+class MetadataLoaderTest extends TestCase
```
nnatter: 🙈

TestCases/Content/ContentDimensionTestCaseTrait.php
```diff
@@ -0,0 +1,77 @@
...
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\ExcerptInterface;
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\SeoInterface;
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\TemplateInterface;
+
+/**
+ * Trait to test your implementation of the ContentDimensionInterface.
+ */
+trait ContentDimensionTestCaseTrait
+{
+    abstract protected function getContentDimensionModel(): ContentDimensionInterface;
```
nnatter: would not suffix this with `Model` because that sounds like a class definition is returned. what do you think about `getContentDimensionInstance`?

TestCases/Content/ContentTestCaseTrait.php
```diff
@@ -0,0 +1,51 @@
...
+
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\ContentDimensionInterface;
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\ContentInterface;
+
+/**
+ * Trait to test your implementation of the ContentInterface.
+ */
+trait ContentTestCaseTrait
+{
+    abstract protected function getModel(): ContentInterface;
```
nnatter: `getContentInstance`?

TestCases/Content/ContentTestCaseTrait.php
```diff
@@ -0,0 +1,51 @@
...
+
+    abstract protected function getModelDimension(int $id): ContentDimensionInterface;
```
nnatter: `getContentDimensionInstance`

TestCases/Content/ExcerptTestCaseTrait.php
```diff
@@ -0,0 +1,130 @@
...
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\ExcerptInterface;
+use Sulu\Bundle\TagBundle\Entity\Tag;
+use Sulu\Bundle\TagBundle\Tag\TagInterface;
+
+/**
+ * Trait to test your implementation of the ExcerptInterface.
+ */
+trait ExcerptTestCaseTrait
+{
+    abstract protected function getExcerptModel(): ExcerptInterface;
```
nnatter: `getExcerptInstace` and so on...

Content/Domain/Model/AbstractContent.php
alexander-schranz: Avoids that somebody need to call this function in constructor.

Content/Infrastructure/Doctrine/MetadataLoader.php
alexander-schranz: ReferenceOption did work but ResolveTargetEntityListener not but can fix this by giving my MetadataLoader a higher priority then the `Doctrine\ORM\Tools\ResolveTargetEntityListener`.

Content/Infrastructure/Doctrine/MetadataLoader.php
alexander-schranz: Not sure if I would go with an intermediate entity because that would make it harder for the entities which use the content bundle but we need to save the `sorting` of the tags somewhere maybe in an own field.

Content/Infrastructure/Doctrine/MetadataLoader.php
alexander-schranz: Confused what you mean?

TestCases/Content/ContentDimensionTestCaseTrait.php
alexander-schranz: Sounds good 👍

Tests/Unit/Content/Infrastructure/Doctrine/MetadataLoaderTest.php
alexander-schranz: Its important magic 😄

Resources/config/services.xml
```diff
@@ -3,6 +3,8 @@
            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
-
+        <service id="sulu_content.doctrine_metadata_loader" class="Sulu\Bundle\ContentBundle\Content\Infrastructure\Doctrine\MetadataLoader">
+            <tag name="doctrine.event_subscriber" priority="100"/>
```
nnatter: maybe add a comment why the priority is needed 🙂

Content/Domain/Model/AbstractContentDimension.php
```diff
@@ -0,0 +1,56 @@
...
+     * @var string
+     */
+    protected $dimensionId;
+
+    public function getId(): int
+    {
+        return $this->id;
+    }
+
+    public function getDimensionId(): string
```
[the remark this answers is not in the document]
alexander-schranz: @nnatter thought about that. I'm not sure if we should handle Locale, Workflow, ... the same how we did handle Template, Seo, Excerpt and so get rid of the whole Dimension entity

Resources/config/services.xml
alexander-schranz: Done

PR #42, opened 2019-11-13
Add doctrine relation between ContentDimension and Dimension

### LINKED ISSUE #39: Create of entity need a flush on add in DimensionRepository
Introduce in https://github.com/sulu/SuluContentBundle/pull/38/files#diff-34aff9a6af073e2504c1c01d0f70b5acR68

The flush of the Dimension shoud be removed when Dimension Domain is moved to Content.

alexander-schranz:
There are cases where we need to access from the `$contentDimension` e.g. the `locale` attributes. That we don't have duplicated data we go here with `$contentDimesion->getDimension()->getLocale()`.

fixes #39

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -24,3 +24,7 @@ parameters:
     autoload_files:
         - vendor/bin/.phpunit/phpunit-8/vendor/autoload.php
         - vendor/autoload.php
+    ignoreErrors:
+        # See https://github.com/phpstan/phpstan/issues/2600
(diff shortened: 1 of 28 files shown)


phpstan.neon
```diff
@@ -24,3 +24,7 @@ parameters:
     autoload_files:
         - vendor/bin/.phpunit/phpunit-8/vendor/autoload.php
         - vendor/autoload.php
+    ignoreErrors:
+        # See https://github.com/phpstan/phpstan/issues/2600
+        - message: '#Parameter \#1 \$x of method Doctrine\\Common\\Collections\\ExpressionBuilder::orX\(\) expects array<int,mixed>|null, Doctrine\\Common\\Collections\\Expr\\Comparison given\.#'
```
alexander-schranz: See https://github.com/phpstan/phpstan/issues/2600 this errors since the new doctrine version

PR #72, opened 2020-01-09
Add possibility to overwrite saveToolbarAction

alexander-schranz:
In somecases you want to replace the SaveToolbarAction with something custom currently it basically impossible to remove or edit toolbarActions as sulu doesn't provide an API for this https://github.com/sulu/sulu/issues/4969.


nnatter:
we might want to override the delete toolbar-action in the future or add an additional action somewhere. this does not work with this solution - would be great if we provide something more flexible here. 

but i cant provide a better approach at the moment, so i would say we go with this for now 🙂

PR #77, opened 2020-02-05
Add preview to entities

luca-rath:
Depends on https://github.com/sulu/sulu/pull/5036

#### What's in this PR?

This pull request adds Preview to custom entities for `content`, `seo` and `excerpt`.

#### Example

To use it, just register the `ContentObjectProvider` for every `$entityClass`, you want to enable it.

```yaml
app.example.preview_object_provider:
    class: Sulu\Bundle\ContentBundle\Content\Infrastructure\Sulu\Preview\ContentObjectProvider
    arguments:
        $entityClass: App\Example\Domain\Model\Example
        $tagFactory: '@sulu_content.tag_factory'
        $categoryFactory: '@sulu_content.category_factory'
    tags:
        - { name: sulu.context, context: admin }
        - { name: sulu_preview.object_provider, provider-key: examples }
```

Note, that the `provider-key` has to match the entity's `resourceKey`!


Content/Infrastructure/Sulu/Admin/ContentViewBuilder.php
```diff
@@ -52,29 +63,38 @@ public function build(
...
+        ToolbarAction $saveToolbarAction,
+        bool $preview
+    ): ViewBuilderInterface {
         $formToolbarActionsWithType = [
             $saveToolbarAction,
             new ToolbarAction(
                 'sulu_admin.type',
                 [
-                    'disabled_condition' => '(_permissions && !_permissions.edit)',
+                    'disable_condition' => '(_permissions && !_permissions.edit)',
```
nnatter: i think this should be `disabled_condition` now 🙂

Content/Infrastructure/Sulu/Admin/ContentViewBuilder.php
```diff
@@ -52,29 +63,38 @@ public function build(
...
     }
 
     protected function buildTemplate(
         string $typeKey,
         string $resourceKey,
         string $parentView,
-        ToolbarAction $saveToolbarAction
-    ): FormViewBuilderInterface {
+        ToolbarAction $saveToolbarAction,
+        bool $preview
```
nnatter: would call these parameters `enablePreview`

Content/Domain/Model/ExcerptInterface.php
```diff
@@ -79,4 +79,11 @@ public function getExcerptIcon(): ?array;
      * @param mixed[]|null $excerptIcon
      */
     public function setExcerptIcon(?array $excerptIcon): void;
+
+    /**
+     * @param mixed[] $data
+     *
+     * @return mixed[]
+     */
+    public function setExcerptData(array $data): array;
```
alexander-schranz: This should basically not exist as the logic is more complex, but at current state we have no way to call the correct services so move that logic Preview Provider where its currently called.

Content/Domain/Model/ExcerptTrait.php
```diff
@@ -215,4 +216,98 @@ private function initializeCategories(): void
             $this->excerptCategories = new ArrayCollection();
         }
     }
+
+    /**
+     * @param mixed[] $data
+     *
+     * @return mixed[]
+     */
+    public function setExcerptData(array $data): array
```
alexander-schranz: This should basically not exist as the logic is more complex, but at current state we have no way to call the correct services so move that logic Preview Provider where its currently called.

Content/Domain/Model/SeoInterface.php
```diff
@@ -42,4 +42,11 @@ public function setSeoNoFollow(bool $seoNoFollow): void;
     public function getSeoHideInSitemap(): bool;
 
     public function setSeoHideInSitemap(bool $seoHideInSitemap): void;
+
+    /**
+     * @param mixed[] $data
+     *
+     * @return mixed[]
+     */
+    public function setSeoData(array $data): array;
```
alexander-schranz: same

Content/Domain/Model/SeoTrait.php
```diff
@@ -119,4 +121,98 @@ public function setSeoHideInSitemap(bool $seoHideInSitemap): void
     {
         $this->seoHideInSitemap = $seoHideInSitemap;
     }
+
+    /**
+     * @param mixed[] $data
+     * 
+     * @return mixed[]
+     */
+    public function setSeoData(array $data): array
```
alexander-schranz: same

Content/Infrastructure/Sulu/Route/ContentRouteDefaultsProvider.php
```diff
@@ -106,7 +110,8 @@ public function isPublished($entityClass, $id, $locale)
 
     public function supports($entityClass)
     {
-        return is_a($entityClass, ContentRichEntityInterface::class, true);
+        return is_a($entityClass, ContentRichEntityInterface::class, true)
+            || is_a($entityClass, ContentProjectionInterface::class, true);
```
alexander-schranz: Is this really needed?

Content/Domain/Model/ExcerptInterface.php
luca-rath: This is just a more complex setter, nothing else. Why is this wrong here?

Content/Infrastructure/Sulu/Route/ContentRouteDefaultsProvider.php
luca-rath: Yes, because the `$object` argument in `getByEntity()` is of type `TemplateInterface`. But the only case where `$object` is not null is preview. So normally, $entity would be of type `ContentProjection`, but for preview, `$entity` would be of type `ContentRichEntity` ...

Content/Domain/Model/ExcerptInterface.php
alexander-schranz: Because mapper and merger should handle that logic, its basically not correct to have the projection here but at current state just move that logic to the preview so we have a working prototype.

luca-rath: Why is it not possible to move it to mapper/merger then? Just curious ...

Tests/Unit/Content/Infrastructure/Sulu/Preview/ContentObjectProviderTest.php
```diff
@@ -0,0 +1,264 @@
...
+
+        $this->contentObjectProvider->setContext($projection, $locale, $context);
+
+        $this->assertSame($context['template'], $projection->getTemplateKey());
+    }
+
+    public function testSerialize()
+    {
+        // Just for Code Coverage
+        $this->contentObjectProvider->serialize(null);
```
alexander-schranz: Could be checked that its a string or stdclass ist correctly returned:

```
$this->assertSame('O:8:"stdClass":0:{}', $this->contentObjectProvider->serialize(new stdclass()));
```

Tests/Unit/Content/Infrastructure/Sulu/Preview/ContentObjectProviderTest.php
```diff
@@ -0,0 +1,264 @@
...
+
+        $this->markTestSkipped();
+    }
+
+    public function testDeserialize()
+    {
+        // Just for Code Coverage
+        $this->contentObjectProvider->deserialize('', null);
```
alexander-schranz: Could check if `O:8:"stdClass":0:{}` isObject.

PR #78, opened 2020-02-12
Symfony update

luca-rath:


alexander-schranz:
@luca-rath we can not update this yet as our dependencies does not allow to update to symfony 5.

luca-rath:
Just wanted to see if tests are passing ...

alexander-schranz:
@luca-rath tests will pass but symfony 5 will not be installed ;)

PR #80, opened 2020-02-21
Use default template if no template key is provided

luca-rath:


luca-rath:
@alexander-schranz is it intended, that PrettyCI fails with those errors? When i run it locally, everything is fine.

PR #97, opened 2020-03-03
Remove unnecessary getId method from DimensionContentInterface

nnatter:


luca-rath:
@nnatter Shouldn't that be removed from `ContentProjection` instead of `DimensionContent`?

PR #99, opened 2020-03-03
Declare entity-independent methods on marker interfaces as static

nnatter:
- [ ] `RoutableInterface:: getContentClass`
- [x] `TemplateInterface:: getTemplateType`
- [ ] `WorkflowInterface:: getWorkflowName`


alexander-schranz:
Making it static will break exist projects code. Should we create a tag maybe before and then provide a UPGRADE path?

alexander-schranz:
The tests are failing because prophecy can not mock static methods.

nnatter:
mocking of static methods is really a pain 🤔 need to think about this..

nnatter:
Replaced by #117

PR #100, opened 2020-03-04
Do not overwrite existing data in TemplateNormalizeEnhancer

nnatter:


nnatter:
@alexander-schranz The `DimensionContent` objects are already merged into a single `ContentProjection` at this point (this is done by the `DataMapper` services).

This change adjusts the serialization/normalization of the `ContentProjection` object. I think that the template-data should not be able to overwrite other data (eg. data of the WorkflowInterface). 

For example, at the moment, if somebody adds a `publishedState` property to his template, it might overwrite the `publishedState` of the `WorkflowInterface` (depending on the order of the executed `NormalizeEnhancer` services).

alexander-schranz:
@nnatter ok know I understand, but should we not throw an exception instead then when a reserved keyword is used?

alexander-schranz:
@nnatter whats your opinion about sulu/sulu#5060. I would really like to have something like that to avoid conflicts between templateData and entity data.

nnatter:
i understand @danrot concerns keeping things in a single place. not sure what would be a good solution here.. maybe we could implement a form-metadata loader that prefixes all paths of content templates or something like that?

PR #101, opened 2020-03-04
Rename RouteDataMapper to RoutableDataMapper to match RoutableInterface, Adjust names of DataMapper test classes

nnatter:


nnatter:
@alexander-schranz do you see any problems with this? should not break backward compatibility if you have not used the `RouteDataMapper` separately

Issue #88, opened 2020-02-25
Provide ContentDataMapper service to call all registered DataMapper services


nnatter:
The services that implement the `DataMapperInterface` are responsible for setting raw data to the correct properties of entities. For example, this is used to set the data that is entered in the Sulu admin interface to the correct properties of the `DimensionContent`.

Right now, these services are located in the `DimensionContentCollectionFactory` namespace. We should move them to a separate namespace because it might make sense to use them independently of the `DimensionContentCollectionFactory`.

I would propose to provide a `ContentDataMapper implements ContentDataMapperInterface` service that contains a collection of all registered `DataMapper` services.

Related:
* https://github.com/sulu/SuluContentBundle/issues/106
* https://github.com/sulu/SuluContentBundle/issues/107

PR #102, opened 2020-03-04
Enhance ContentObjectProvider

luca-rath:
This adds a bc break, because the constructor of `ContentObjectProvider` has changed.

### DIFF
--- a/Content/Infrastructure/Sulu/Preview/ContentObjectProvider.php
+++ b/Content/Infrastructure/Sulu/Preview/ContentObjectProvider.php
@@ -15,10 +15,11 @@ namespace Sulu\Bundle\ContentBundle\Content\Infrastructure\Sulu\Preview;
 
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\NoResultException;
+use Sulu\Bundle\ContentBundle\Content\Application\ContentDataMapper\DataMapper\ExcerptDataMapper;
+use Sulu\Bundle\ContentBundle\Content\Application\ContentDataMapper\DataMapper\SeoDataMapper;
+use Sulu\Bundle\ContentBundle\Content\Application\ContentDataMapper\DataMapper\TemplateDataMapper;
 use Sulu\Bundle\ContentBundle\Content\Application\ContentResolver\ContentResolverInterface;
 use Sulu\Bundle\ContentBundle\Content\Domain\Exception\ContentNotFoundException;
-use Sulu\Bundle\ContentBundle\Content\Domain\Factory\CategoryFactoryInterface;
-use Sulu\Bundle\ContentBundle\Content\Domain\Factory\TagFactoryInterface;
 use Sulu\Bundle\ContentBundle\Content\Domain\Model\ContentProjectionInterface;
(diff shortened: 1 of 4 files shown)


Content/Infrastructure/Sulu/Preview/ContentObjectProvider.php
```diff
@@ -124,196 +131,18 @@ public function getId($object)
     public function setValues($object, $locale, array $data)
     {
         if ($object instanceof SeoInterface) {
-            $this->setSeoData($object, $data);
+            $this->seoDataMapper->map($data, $object, $object);
```
nnatter: would it make sense to call all registered mappers like in the `DimensionContentCollectionFactory`? we already discussed about making them usable in other places in https://github.com/sulu/SuluContentBundle/issues/88

luca-rath: @alexander-schranz what do you think about this?

alexander-schranz: @nnatter @luca-rath not sure if we should only create a Registry or have a single service  DataMapper service with the `map` function which calls the subservices. So he don't need todo everytime a foreach just call the DataMapper service.

No matter what we chose we should do the same for the *`Mergers`*.

nnatter: yes, i think we should provide a service that automatically calls all registered `DataMapper`s instead of a registry. but what name should we use? 🙈

alexander-schranz: @nnatter DataMapperInterface and DataMapperEnhancerInterface. I don't like the Enhancer postfix but atleast we would be consistent. Same for `MergerInterface` and `MergerEnhancerInterface`. What do you think?

nnatter: i think i would go with a `Behaviour` prefix. this would lead to a `DataMapperInterface` that manages a collection of `BehaviourDataMapperInterface` objects.

alternatively i would also be okay with a `ContentDataMapperInterface` that has `DataMapperInterface` objects. That way, we dont have to rename anything existant 🙂

alexander-schranz: @nnatter ContentDataMapper sounds laso good at current state for me. I think the sub services should also be in its own namespace like we have it for the enhancers e.g.

Application
 - ContentDataMapper
    - ContentDataMapper.php
    - ContentDataMapperInterface.php
    - DataMapper
      - DataMapperInterface.php
      - ExcerptDataMapper.php
      - SeoDataMapper.php
      - TemplateDataMapper.php

@nnatter what do you think?

Content/Infrastructure/Sulu/Preview/ContentObjectProvider.php
```diff
@@ -72,16 +63,14 @@ public function __construct(
         StructureMetadataFactoryInterface $structureMetadataFactory,
```
nnatter: this is never used in this class right? 🤔

PR #103, opened 2020-03-04
Remove unnecessary default_type configuration in ExampleTestBundle

nnatter:


alexander-schranz:
@nnatter I don't think we can remove this. Currently I think it only works because of a bug in sulu see sulu/sulu#5055

nnatter:
@alexander-schranz i think Sulu should not throw an exception but use the first template if no `default_type` is set. would propose to keep this pull request open until the Sulu issue is fixed. then we can discuss this again 🙂

alexander-schranz:
Could not find a way to fix this currently in sulu/sulu. The structures in sulu/sulu are loaded async and not in a compilerpass so its not possible to set the `default_type` in a compilerpass. As we need this at some places e.g. when no template is given I will not remove it from the example as it recommended to be defined correctly.

Issue #104, opened 2020-03-06
Remove abstract classes as they dont provide any value


nnatter:
The implementation of the `AbstractContentRichEntity` class is:
```php
abstract class AbstractContentRichEntity implements ContentRichEntityInterface
{
    use ContentRichEntityTrait;
}
```

The implementations of the `AbstractDimensionContent` and `AbstractContentProjection` look very similar. I think these classes make things harder to understand/confusing while not providing any value.

If the user implements his own custom entity, he will ask himself if he needs to extend the `AbstractContentRichEntity` class or if it is enough to extend the `ContentRichEntityInterface`. Furthermore, the class contains very few code and it is not a problem to just use the `ContentRichEntityTrait` in the custom entity.

Therefore, I think we should remove the following classes from the bundle:
* `AbstractContentProjection`
* `AbstractContentRichEntity`
* `AbstractDimensionContent`

alexander-schranz:
fixed in #105

PR #105, opened 2020-03-06
Remove abstract classes that dont provide any value

nnatter:
See #104


alexander-schranz:
@nnatter it was discussed some time ago that the AbstractClasses should implement all behaviours which are provided by the Sulu Content Bundle. So if I have a entity which should have full content capability it should extend the abstract class.

alexander-schranz:
@nnatter created an issue for my comment #112 so we can merge this.

PR #108, opened 2020-03-06
Add priority to DataMapper, Merger and NormalizeEnhancer service definitions

nnatter:
This makes it possible to influence the position of a custom `DataMapper`/`Merger`/`NormalizeEnhancer`

### DIFF
--- a/Resources/config/mapper.xml
+++ b/Resources/config/mapper.xml
@@ -8,18 +8,18 @@
             <argument type="service" id="sulu_page.structure.factory"/>
             <argument>%sulu.content.structure.default_types%</argument>
 
-            <tag name="sulu_content.data_mapper"/>
+            <tag name="sulu_content.data_mapper" priority="3072"/>
         </service>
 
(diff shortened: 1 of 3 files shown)


Resources/config/mapper.xml
```diff
@@ -8,18 +8,18 @@
             <argument type="service" id="sulu_page.structure.factory"/>
             <argument>%sulu.content.structure.default_types%</argument>
 
-            <tag name="sulu_content.data_mapper"/>
+            <tag name="sulu_content.data_mapper" priority="3072"/>
```
alexander-schranz: I don't understand the usage of this high numbers.

nnatter: i dont care about the specific numbers, just used the same pattern that sulu uses 🙂 what would you use?

alexander-schranz: I think we should follow the rule what symfony does they have internal priorites only between `-255` to `255`.

Would go in this case with:

Template: `128` 
Workflow: `64` 
Excerpt: `32` 
Seo: `16` 
Routeable: `-32`

Issue #107, opened 2020-03-06
Rename ContentProjectionNormalizer to ContentNormalizer


nnatter:
The `ContentProjectionNormalizer` is responsible for transforming a `ContentProjection` entity into a normalized array. For example, this is used for generating the data array that is sent as `json` response to the Form view of the Sulu admin.

As we want to make the `ContentProjection` entity optional (see https://github.com/sulu/SuluContentBundle/issues/82), we should rename the `ContentProjectionNormalizer` service to `ContentNormalizer` and adjust it to **accept not only `ContentProjection` entities but every object** as argument.

Furthermore, i would propose to rename the `NormalizeEnhancerInterface` to `NormalizerInterface` to keep things consistent with https://github.com/sulu/SuluContentBundle/issues/106 and https://github.com/sulu/SuluContentBundle/issues/82 .

Related:
* https://github.com/sulu/SuluContentBundle/issues/106
* https://github.com/sulu/SuluContentBundle/issues/88
* https://github.com/sulu/SuluContentBundle/issues/82

nnatter:
Partially fixed by https://github.com/sulu/SuluContentBundle/pull/111

Renaming of the `NormalizeEnhancerInterface` is not done yet.

PR #111, opened 2020-03-08
Rename ContentProjectionNormalizer to ContentNormalizer

alexander-schranz:
The normalizer should not depend on the ContentProjectionInterface, it should work with every contentable object.


alexander-schranz:
Did see #107 just know. You want to include `Data` in the service name?

nnatter:
i think i would either include `Data` here or remove it from the `DataMapper` classes. what do you think?

alexander-schranz:
@nnatter thinking about it I'm not sure if its the same as `ContentNormalizer` does not normalize data it does normalize a object and the output is data, so I personally would more go with this way or removing it also from Mapper namespaces.

nnatter:
yes, you are right 🙂 sounds good

PR #115, opened 2020-03-09
Declare getTemplateType method on TemplateInterface as static

nnatter:
Partially replaces #99


alexander-schranz:
Feels strange but think this is the only way currently todo this and it only affects us when writing tests.

PR #116, opened 2020-03-10
Declare getWorkflowName method on WorkflowInterface as static

nnatter:
Partially replaces #99
Builds upon: https://github.com/sulu/SuluContentBundle/pull/115


nnatter:
Replaced by #117

PR #122, opened 2020-03-19
Update phpstan dependencies to version 0.12

nnatter:


Content/Infrastructure/Doctrine/DimensionRepository.php
```diff
@@ -89,7 +89,7 @@ public function findOneBy(array $criteria): ?DimensionInterface
 
     public function findBy(array $criteria): iterable
     {
-        /** @var DimensionInterface[] $directory */
+        /** @var DimensionInterface[] $directories */
```
alexander-schranz: @nnatter 😄 think this should be called `$dimensions`

PR #123, opened 2020-03-19
Rename parameters of MergerInterfaceto not include ContentProjection

nnatter:


alexander-schranz:
@nnatter should we maybe `Swap` the parameter now 🤔

nnatter:
dont have an opinion on that 🙁 would you like it better if they were swapped?

alexander-schranz:
@nnatter feels more natural for me but we can discuss this here: https://github.com/sulu/SuluContentBundle/issues/124

PR #127, opened 2020-03-20
Move code related to ContentProjection from ContentNormalizer to ContentProjectionNormalizer

nnatter:
Did not think about this when implementing https://github.com/sulu/SuluContentBundle/issues/107


alexander-schranz:
Also when using DimensionContent object the id will mit represent the correct id, but at current state its im for me :)

alexander-schranz:
@nnatter tests are failing

nnatter:
i think we will handle the `DimensionContent` case when removing the `ContentProjection` 🙂

PR #128, opened 2020-03-23
Replace DimensionId with Dimension in ContentProjection

luca-rath:


Tests/Functional/Integration/responses/example_get.json
```diff
@@ -1,5 +1,12 @@
 {
-    "dimensionId": "@uuid@",
+    "dimension": {
+        "locale": "en",
+        "stage": "draft",
+        "attributes": {
+            "locale": "en",
+            "stage": "draft"
+        }
+    },
```
luca-rath: I have no idea, why the id isn't being serialized?? @nnatter

nnatter: because the `ContentProjectionNormalizer` adds it as ignored-attribute. i think i would adjust that class to completely omit all information about the `dimension` in the normalized array

alexander-schranz:
Don't like this change but hopefully we will remove the ContentProjectionInterface soon.

nnatter:
> Don't like this change but hopefully we will remove the ContentProjectionInterface soon.

Yes, the idea is to not waste any resources here at the moment because we are going to remove the entity soon anyway.

Tests/Functional/Integration/responses/example_get.json
alexander-schranz: we could add a `getDimensionId()` to the trait and add `dimension` as ignored property for ContentProjectionInterface.

nnatter: do you see any value in serializing the `dimensionId`?

luca-rath: fixed without adding dimensionId

Tests/Unit/Content/Infrastructure/Sulu/Route/ContentRouteDefaultsProviderTest.php
```diff
@@ -78,13 +78,16 @@ public function testIsPublished(): void
...
-        $dimensionRepository->findOneBy(['id' => '123-456'])->willReturn(new Dimension('123-456', [
+        $dimension = new Dimension('123-456', [
             'locale' => 'en',
             'stage' => 'live',
-        ]));
+        ]);
+
+        $contentProjection->getDimension()->willReturn($dimension);
+
+        $dimensionRepository = $this->prophesize(DimensionRepositoryInterface::class);
```
nnatter: should not be needed anymore. will fix this by myself 🙂

Issue #82, opened 2020-02-21
Make ContentProjection optional and use it only when rendering the Website


alexander-schranz:
Right now, every `ContentRichEntity` needs a `DimensionContent` and a `ContentProjection` class. In most cases, these classes look very similar. Furthermore, a `ContentProjection` is instantiated when resolving the content for specific `dimensionAttributes`, even in the context of the admin.

We should adjust the bundle to make the implementation of a `ContentProjection` optional. Furthermore, we should never instantiate a `ContentProjection` in the admin context. The `ContentProjection` should only act as view-model in the website context to improve the performance.

When implementing this, we should adjust all `Merger` classes to not expect a `ContentProjection` instance but accept all objects as `target`.

alexander-schranz:
The ContentProjection was removed in #133

PR #133, opened 2020-04-27
Remove ContentProjection from bundle source code

nnatter:
This PR removes the `ContentProjection` concept from the bundle. In most places, the `ContentProjection` was plainly replaced with the `DimensionContent`. 

The PR adds a `mergeCollection` method to the `ContentMerger` as replacement for the `ContentProjectionFactory`. Furthermore, the PR renames the `ContentCopierInterface:: copyFromContentProjection` method to `copyFromDimensionContent`.

See: https://github.com/sulu/SuluContentBundle/issues/82


Content/Application/ContentPersister/ContentPersister.php
```diff
@@ -63,6 +63,6 @@ public function persist(ContentRichEntityInterface $contentRichEntity, array $da
             $data
         );
 
-        return $this->viewFactory->create($dimensionContentCollection);
+        return $this->contentMerger->mergeCollection($dimensionContentCollection);
```
nnatter, line 66: what do you think about the `mergeCollection()` method on the `ContentMerger`?

Content/Infrastructure/Sulu/Route/ContentRouteDefaultsProvider.php
```diff
@@ -103,7 +103,7 @@ public function isPublished($entityClass, $id, $locale)
     public function supports($entityClass)
     {
         return is_a($entityClass, ContentRichEntityInterface::class, true)
-            || is_a($entityClass, ContentProjectionInterface::class, true);
+            || is_a($entityClass, DimensionContentInterface::class, true);
```
nnatter, line 106: i guess this had something to do with the preview - do you know if this is still needed? @luca-rath

luca-rath: I think this is still needed, because the `ContentObjectProvider::deserialize` will return a `DimensionContent` instead of a `ContentRichEntity`.

Content/Application/ContentMerger/ContentMergerInterface.php
```diff
@@ -13,7 +13,12 @@
 
 namespace Sulu\Bundle\ContentBundle\Content\Application\ContentMerger;
 
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\DimensionContentCollectionInterface;
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\DimensionContentInterface;
+
 interface ContentMergerInterface
 {
+    public function mergeCollection(DimensionContentCollectionInterface $dimensionContentCollection): DimensionContentInterface;
```
nnatter, line 21: what do you think about the `mergeCollection()` method on the `ContentMerger` as replacement for the `ContentProjectionFactory`?

alexander-schranz: @nnatter I don't see why we should have 2 functions in this interface? I would just give the `merge` the `collection` at current state.

Content/Domain/Model/DimensionContentInterface.php
```diff
@@ -17,5 +17,9 @@ interface DimensionContentInterface
 {
     public function getDimension(): DimensionInterface;
 
-    public function createProjectionInstance(): ContentProjectionInterface;
+    public function getContentRichEntity(): ContentRichEntityInterface;
```
alexander-schranz, line 20: do we need this in the interface?

nnatter: if we want to reuse the `createDimensionContent()` to create the merged dimension, i would say so 🙂 but i think this could be helpful in otherscenarioss too

https://github.com/sulu/SuluContentBundle/pull/133/files#diff-de2546549df55e9685857d68dfcd0e95R54-R57

Content/Application/ContentMerger/ContentMergerInterface.php
nnatter: that way it is harder to call all the mergers for a specific object. but as far as i remember, we did not have to do that anywhere. so that would be okay for me too 🙂

alexander-schranz: I would keep every interface as small as possible. Maybe best would just we keep the ContentProjectionFactory and don't touch the merger and the Factory just return a merged `DimensionContentInterface` what do you think about that?

alexander-schranz: As discussed over slack we keep go at current state with `merge(DimensionContentCollectionInterface $collection): DimensionContentInterface`

wachterjohannes: its ok for me - we could introduce a `DimensionContentCollectionInterface::merge` method once - but then we would have to inject the mergers into the constructor.

Content/Domain/Model/DimensionContentInterface.php
alexander-schranz: Ok not sure about it but its excluded from serialization so its ok.

PR #135, opened 2020-04-29
Refactor content-view builder

### LINKED ISSUE #98: Refactor ContentViewBuilder to accept ContentRichtEntity class as argument
Right now, the public interface of the `ContentViewBuilder` contains only the following method:

```php
    public function build(
        ViewCollection $viewCollection,
        string $resourceKey,
        string $typeKey,
        string $editParentView,
        string $addParentView = null,
        ?ToolbarAction $saveToolbarAction = null
    ): void;
```

This is impractical because of various reasons. Including that we always generate all views (template, seo, excerpt), we need to pass the `resourceKey` explicitly and it is hard to overwrite/configure toolbar actions. Also, the name of the class itself is confusing and not consistent with the `ViewBuilder` concept of Sulu.

To solve these issues, I propose the following changes:
* Rename the class to `ContentViewBuilderFactory` to match the `ViewBuilderFactory` of Sulu. Furthermore adjust the methods of the class to return the `ViewBuilder` instances instead of adding them to a `ViewCollection` that needs to be passed as argument.
* Pass a `ContentRichtEntity` class instead of a `$resourceKey` argument and a `$typeKey`.  The bundle checks if an object implements a specific interface in a lot of places. We can reuse this pattern here via the `is_subclass_of` interface. For example, create a SEO view only if the class implements the `SeoInterface`. Furthermore, we can utilize the static `getResourceKey` method (which is not used anywhere right now) of the `ContentRichEntityInterface` to gather the `$resourceKey`. A similar static method should be introduced on the `TemplateInterface`.
* There might be usecases in which we need to create only specific views and dont want to create all views for the implemented interfaces. For this cases, we should expose the methods for creating single views. Also, this views should require the concrete values as parameter to allow for creating views without a `ContentRichtEntity` class.
* It is not useful to only provide a `$saveToolbarAction` parameter. There are usecases in which we want to overwrite the delete button or even omit one of the toolbar actions for the created views. Furthermore, i dont think it is sustainable to add a parameter for each toolbar action. Therefore i would propose to provide a nullable `$toolbarActions` parameter on the methods that allow to create single views.

These adjustments result in a public interface that looks something like:
```php
    public function createViews(
        string $entityClass,
        string $editParentView,
        string $addParentView = null
    ): array;

    public function createTemplateFormView(
        string $typeKey,
        string $resourceKey,
        string $parentView,
        ?array $toolbarActions = null
    ): array;

    // ...
```

wachterjohannes:
This PR refactors the ContentViewBuilder to match the requirements of the issue #98 

Fixes #98


Content/Infrastructure/Sulu/Admin/ContentViewBuilderFactory.php
```diff
@@ -0,0 +1,260 @@
...
+                        ]
+                    ),
+                ]
+            ),
+        ];
+    }
+
+    public function createContentRichViews(
+        string $entityClass,
+        string $templateFormKey,
```
wachterjohannes, line 109: use static method

Content/Infrastructure/Sulu/Admin/ContentViewBuilderFactory.php
```diff
@@ -0,0 +1,264 @@
...
+                        [
+                            'visible_condition' => '(!_permissions || _permissions.live)',
+                        ]
+                    ),
+                ]
+            ),
+        ];
+    }
+
+    public function createContentRichViews(
```
alexander-schranz, line 107: Would not go on the rich just call it `createViews`

Content/Infrastructure/Sulu/Admin/ContentViewBuilderFactory.php
```diff
@@ -0,0 +1,264 @@
...
+        string $entityClass,
+        string $editParentView,
+        ?string $addParentView = null,
+        ?array $toolbarActions = null,
```
alexander-schranz, line 111: make `toolbarActions` the last parameter

Content/Infrastructure/Sulu/Admin/ContentViewBuilderFactory.php
```diff
@@ -0,0 +1,264 @@
...
+
+        if (!$this->hasPermission($securityContext, PermissionTypes::DELETE)) {
+            unset($toolbarActions['delete']);
+            unset($addToolbarActions['delete']);
+        }
+
+        $views = [];
+
+        if ($this->hasPermission($securityContext, PermissionTypes::ADD)) {
+            if ($addParentView && is_subclass_of($dimensionContentClass, TemplateInterface::class)) {
```
alexander-schranz, line 150: nesting not needed use `&&`

Content/Infrastructure/Sulu/Admin/ContentViewBuilderFactory.php
```diff
@@ -0,0 +1,264 @@
...
+                    $resourceKey,
+                    $seoAndExcerptToolbarActions
+                );
+            }
+        }
+
+        return $views;
+    }
+
+    public function createTemplateFormView(
```
alexander-schranz, line 194: `make private` they were only here because of the detection not implemented yet

Content/Infrastructure/Sulu/Admin/ContentViewBuilderFactory.php
```diff
@@ -0,0 +1,264 @@
...
+        return $this->createFormViewBuilder($parentView . '.content', '/content', $previewEnabled)
+            ->setResourceKey($resourceKey)
+            ->setFormKey($formKey)
+            ->setTabTitle('sulu_content.content')
+            ->addToolbarActions(array_values($toolbarActions ?: $this->getDefaultToolbarActions()))
+            ->setTabOrder(20)
+            ->setParent($parentView);
+    }
+
+    public function createSeoFormView(
```
alexander-schranz, line 210: `make private` they were only here because of the detection not implemented yet

Content/Infrastructure/Sulu/Admin/ContentViewBuilderFactory.php
```diff
@@ -0,0 +1,264 @@
...
+            ->setResourceKey($resourceKey)
+            ->setFormKey('content_seo')
+            ->setTabTitle('sulu_content.seo')
+            ->setTitleVisible(true)
+            ->addToolbarActions(array_values($toolbarActions ?: [$this->getDefaultToolbarActions()['save']]))
+            ->setTabOrder(30)
+            ->setParent($parentView);
+    }
+
+    public function createExcerptFormView(
```
alexander-schranz, line 226: `make private` they were only here because of the detection not implemented yet

Content/Infrastructure/Sulu/Admin/ContentViewBuilderFactory.php
```diff
@@ -0,0 +1,264 @@
...
+            ->setTitleVisible(true)
+            ->addToolbarActions(array_values($toolbarActions ?: [$this->getDefaultToolbarActions()['save']]))
+            ->setTabOrder(40)
+            ->setParent($parentView);
+    }
+
+    /**
+     * @return PreviewFormViewBuilderInterface|FormViewBuilderInterface
+     */
+    protected function createFormViewBuilder(string $name, string $path, bool $previewEnabled): ViewBuilderInterface
```
alexander-schranz, line 245: make private

Content/Infrastructure/Sulu/Admin/ContentViewBuilderFactory.php
```diff
@@ -0,0 +1,264 @@
...
+        if ($previewEnabled) {
+            $formViewBuilder = $this->viewBuilderFactory->createPreviewFormViewBuilder($name, $path);
+        } else {
+            $formViewBuilder = $this->viewBuilderFactory->createFormViewBuilder($name, $path);
+        }
+
+        return $formViewBuilder;
+    }
+
+    protected function hasPermission(?string $securityContext, string $permissionType): bool
```
alexander-schranz, line 256: make private

Content/Infrastructure/Sulu/Admin/ContentViewBuilderFactory.php
```diff
@@ -0,0 +1,264 @@
...
+            ),
+        ];
+    }
+
+    public function createContentRichViews(
+        string $entityClass,
+        string $editParentView,
+        ?string $addParentView = null,
+        ?array $toolbarActions = null,
+        ?string $securityContext = null
```
alexander-schranz, line 112:
```php
    public function createContentRichViews(
        string $entityClass,
        string $editParentView,
        ?string $addParentView = null,
        ?string $securityContext = null,
        ?array $toolbarActions = null
    ): array {
```

Content/Infrastructure/Sulu/Admin/ContentViewBuilderFactoryInterface.php
```diff
@@ -0,0 +1,69 @@
...
+        string $editParentView,
+        ?string $addParentView = null,
+        ?array $toolbarActions = null,
+        ?string $securityContext = null
+    ): array;
+
+    /**
+     * @param array<string, ToolbarAction> $toolbarActions
+     */
+    public function createTemplateFormView(
```
alexander-schranz, line 42: remove from interface

Content/Infrastructure/Sulu/Admin/ContentViewBuilderFactoryInterface.php
```diff
@@ -0,0 +1,69 @@
...
+        bool $previewEnabled,
+        string $resourceKey,
+        string $formKey,
+        ?array $toolbarActions = null
+    ): ViewBuilderInterface;
+
+    /**
+     * @param array<string, ToolbarAction> $toolbarActions
+     */
+    public function createSeoFormView(
```
alexander-schranz, line 53: remove from interface

Content/Infrastructure/Sulu/Admin/ContentViewBuilderFactoryInterface.php
```diff
@@ -0,0 +1,69 @@
...
+        string $parentView,
+        bool $previewEnabled,
+        string $resourceKey,
+        ?array $toolbarActions = null
+    ): ViewBuilderInterface;
+
+    /**
+     * @param array<string, ToolbarAction> $toolbarActions
+     */
+    public function createExcerptFormView(
```
alexander-schranz, line 63: remove from interface

Content/Infrastructure/Sulu/Admin/ContentViewBuilderFactory.php
```diff
@@ -0,0 +1,264 @@
...
+
+        /** @var callable $callable */
+        $callable = [$dimensionContentClass, 'getTemplateType'];
+        $templateFormKey = \call_user_func($callable);
+
+        $previewEnabled = $this->objectProviderRegistry->hasPreviewObjectProvider($resourceKey);
+
+        $toolbarActions = $toolbarActions ?: $this->getDefaultToolbarActions();
+        $addToolbarActions = $toolbarActions;
+        $seoAndExcerptToolbarActions = ['save' => $toolbarActions['save']];
```
alexander-schranz, line 130: `save` could not be set and so need to be checked if exists:

```
$seoAndExcerptToolbarActions = [];
if (isset($toolbarActions['save'])) {
    $seoAndExcerptToolbarActions = ['save' => $toolbarActions['save']];
}
```

for the unset this is not important as unset dont throw error when not set.

UPGRADE.md
```diff
@@ -2,14 +2,48 @@
...
+### Refactored the ContentViewBuilder
+
+The class and its interface was renamed from `ContentViewBuilder` & `ContentViewBuilderInterface` 
+to `ContentViewBuilderFactory` & `ContentViewBuilderFactoryInterface`.
+
+The service has been renamed from `sulu_content.content_view_builder` to `sulu_content.content_view_builder_factory`.
+
+The function `build` was replaced by `createContentRichViews` and additional functions has been introduced.
+
+__BEFORE:__
```
alexander-schranz, line 14: add empty line after it and use `**before**:` like we do in sulu upgrade

```
**before**:

CODE

**after**:

CODE
```

UPGRADE.md
```diff
@@ -2,14 +2,48 @@
...
+$this->contentViewBuilder->build(
+    $viewCollection,
+    $resourceKey,
+    Example::TEMPLATE_TYPE,
+    static::EDIT_TABS_VIEW,
+    static::ADD_TABS_VIEW
+);
+```
+
+__AFTER:__
```
alexander-schranz, line 25: empty line after (some MD reader have problems if no empty line appears before code block) it and use `**after**:`

UPGRADE.md
```diff
@@ -2,14 +2,48 @@
...
 
 The arguments of the constructor of the `ContentObjectProvider` class were changed.
 
-### Renamed sulu_content.content_projection_factory_merger tag to sulu_content.merger
+### Renamed the `sulu_content.content_projection_factory_merger` tag to `sulu_content.merger`
 
 The `sulu_content.content_projection_factory_merger` tag was renamed to `sulu_content.merger`.
 
-### Renamed sulu_content.normalize_enhancer tag to sulu_content.normalizer
+### Renamed the `sulu_content.normalize_enhancer` tag to `sulu_content.normalizer`
```
alexander-schranz, line 47: 👍

UPGRADE.md
```diff
@@ -2,14 +2,48 @@
...
 
-### Changed constructor of ContentObjectProvider class
+### Refactored the ContentViewBuilder
+
+The class and its interface was renamed from `ContentViewBuilder` & `ContentViewBuilderInterface` 
+to `ContentViewBuilderFactory` & `ContentViewBuilderFactoryInterface`.
+
+The service has been renamed from `sulu_content.content_view_builder` to `sulu_content.content_view_builder_factory`.
+
+The function `build` was replaced by `createContentRichViews` and additional functions has been introduced.
```
alexander-schranz, line 12: `createViews`.

Add line that the specific template excerpt and seo are are removed and the entity is now autodetected.

Content/Infrastructure/Sulu/Admin/ContentViewBuilderFactory.php
wachterjohannes: in my opinion i would keep this functions public - does not hurt that they are there. in the issue we also defined it as public and there are maybe usecases to create them manually

alexander-schranz: No I want to keep the public interface small and don't want to fight with BC when changing this functions.

Issue #92, opened 2020-02-25
Set the current workflowPlace of the ContentWorkflow only on a specific DimensionContent


nnatter:
The `ContentWorkflow` service allows to apply a transition on a `contentRichEntity` and given `dimensionAttributes`. Because the `currentPlace` of an `contentRichEntity` might differ based on the locale, the `currentPlace` is stored on the `DimensionContent` entity. This makes sense, as the english version of a page might be published while the german version is still in review.

In most setups, there are 4 `DimensionContent` entities that are related to a specific locale:
* Draft Stage / Unlocalized
* Draft Stage / Localized
* Live Stage / Unlocalized
* Live Stage / Localized

Right now, all of these `DimensionContent` entities contain a `currentPlace` property, but the ContentWorkflow only updates the `currentPlace` property of the `Draft Stage / Localized` `DimensionContent`. This means the database will contain the following values for the `currentPlace` attributes, if a `contentRichEntity` is published in a specific locale:
* Draft Stage / Unlocalized: `unpublished`
* Draft Stage / Localized: `published`
* Live Stage / Unlocalized: `unpublished`
* Live Stage / Localized: `unpublished`

In my opinion, it is very confusing that the `Draft Stage / Localized` `DimensionContent` is the only entity that contains the correct `currentPlace`. We should adjust this behaviour to one of the following strategies:

* Update the `currentPlace` of all entities. This is problematic because there is no "right" value for the unlocalized `DimensionContent`
* Set the `currentPlace` only on the `Draft Stage / Localized` `DimensionContent`. Set the `currentPlace` of all other `DimensionContent` entities to `null`

wachterjohannes:
in my opinion it should look like this:

* Draft Stage / Unlocalized: null
* Draft Stage / Localized: published
* Live Stage / Unlocalized: null
* Live Stage / Localized: published

nnatter:
After thinking through this a second time, I have the opinion that we should save the `currentPlace` **only on the `Draft Stage / Localized``DimensionContent`** and set the `currentPlace` of all other `DimensionContent` entities to `null`.

luca-rath:
It's necessary to have `workflowPublished` on the `Live Stage / Localized` `DimensionContent` to allow sorting by published date.

I'm using following code to evaluate `publishedState` of a `DimensionContent`.
That code should work with both, `DRAFT` and `LIVE` dimensions.

* `getPublishedState() === true` means `published`
[...]

PR #138, opened 2020-05-06
Copy published date to live DimensionContent

luca-rath:
Related to #92

### DIFF
--- a/Content/Application/ContentDataMapper/DataMapper/WorkflowDataMapper.php
+++ b/Content/Application/ContentDataMapper/DataMapper/WorkflowDataMapper.php
@@ -1,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\ContentBundle\Content\Application\ContentDataMapper\DataMapper;
+
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\DimensionContentInterface;
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\DimensionInterface;
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\WorkflowInterface;
+
+class WorkflowDataMapper implements DataMapperInterface
+{
+    public function map(
+        array $data,
+        object $unlocalizedObject,
+        ?object $localizedObject = null
+    ): void {
+        if (!$localizedObject
+            || !$localizedObject instanceof WorkflowInterface
+            || !$localizedObject instanceof DimensionContentInterface) {
+            return;
+        }
+
+        if (DimensionInterface::STAGE_LIVE !== $localizedObject->getDimension()->getStage()) {
+            return;
+        }
+
+        $published = $data['published'] ?? null;
+
+        if (!$published) {
+            return;
+        }
+
+        $localizedObject->setWorkflowPublished(new \DateTimeImmutable($published));
+    }
+}
--- a/Content/Application/ContentDataMapper/DataMapper/WorkflowDataMapper.php
+++ b/Content/Application/ContentDataMapper/DataMapper/WorkflowDataMapper.php
@@ -24,13 +24,30 @@ class WorkflowDataMapper implements DataMapperInterface
         object $unlocalizedObject,
         ?object $localizedObject = null
     ): void {
-        if (!$localizedObject
-            || !$localizedObject instanceof WorkflowInterface
-            || !$localizedObject instanceof DimensionContentInterface) {
+        if (!$unlocalizedObject instanceof WorkflowInterface) {
             return;
         }
 
-        if (DimensionInterface::STAGE_LIVE !== $localizedObject->getDimension()->getStage()) {
+        if ($localizedObject) {
+            if (!$localizedObject instanceof WorkflowInterface) {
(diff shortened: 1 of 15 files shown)


luca-rath:
@nnatter @alexander-schranz What do you think about this solution?

Content/Application/ContentDataMapper/DataMapper/WorkflowDataMapper.php
```diff
@@ -0,0 +1,45 @@
...
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\WorkflowInterface;
+
+class WorkflowDataMapper implements DataMapperInterface
+{
+    public function map(
+        array $data,
+        object $unlocalizedObject,
+        ?object $localizedObject = null
+    ): void {
+        if (!$localizedObject
```
nnatter, line 27: should we add similar checks as in the other `DataMapper` classes?

wachterjohannes:
i have not looked at the code in detail - but we already has an issue which could be related: https://github.com/sulu/SuluContentBundle/issues/92

luca-rath:
> i have not looked at the code in detail - but we already has an issue which could be related: #92

I already linked that issue in the pr description :)

Content/Application/ContentDataMapper/DataMapper/WorkflowDataMapper.php
```diff
@@ -0,0 +1,45 @@
...
+            || !$localizedObject instanceof WorkflowInterface
+            || !$localizedObject instanceof DimensionContentInterface) {
+            return;
+        }
+
+        if (DimensionInterface::STAGE_LIVE !== $localizedObject->getDimension()->getStage()) {
+            return;
+        }
+
+        $published = $data['published'] ?? null;
```
nnatter, line 37: is `$data['published']` set on the first publish transition? not sure if doctrine can handle this correctly

Content/Application/ContentDataMapper/DataMapper/WorkflowDataMapper.php
```diff
@@ -0,0 +1,45 @@
...
+        object $unlocalizedObject,
+        ?object $localizedObject = null
+    ): void {
+        if (!$localizedObject
+            || !$localizedObject instanceof WorkflowInterface
+            || !$localizedObject instanceof DimensionContentInterface) {
+            return;
+        }
+
+        if (DimensionInterface::STAGE_LIVE !== $localizedObject->getDimension()->getStage()) {
```
nnatter, line 33: this feels kind of hacky. but the magic `setWorkflowPlace` felt hacky too, so i think its a good exchange 🙂

Content/Application/ContentWorkflow/Subscriber/PublishTransitionSubscriber.php
```diff
@@ -41,6 +42,12 @@ public function onPublish(TransitionEvent $transitionEvent): void
             return;
         }
 
+        if ($dimensionContent instanceof WorkflowInterface) {
```
nnatter, line 45: not sure if doctrine can handle this in a way that the `contentCopier` will know about this change

Content/Application/ContentDataMapper/DataMapper/WorkflowDataMapper.php
luca-rath: I wouldn't, because i guess, it should be possible to inject e.g. the `SeoDataMapper` in your project and just use it with one DimensionContent.

Content/Application/ContentDataMapper/DataMapper/WorkflowDataMapper.php
luca-rath: yes it is

Content/Application/ContentWorkflow/Subscriber/PublishTransitionSubscriber.php
luca-rath: yes it can, because here the dimensionContent is being changed, that is included in the DimensionContentCollection that is being passed to the ContentCopier

Content/Application/ContentDataMapper/DataMapper/WorkflowDataMapper.php
nnatter: well the idea behind these checks was that it should be possible to handle unlocalized entities that have no localized dimension at all. so if we want to do that we need a similar check here right? 🙂

luca-rath: Ah, now i understand the question 😄 Ok, gonna change that

luca-rath: fixed

nnatter:
looks good to me 👍 curious what @alexander-schranz thinks 🙂

alexander-schranz:
@nnatter @luca-rath Not sure if I would relly on it that the ContentCopier will use the persister and so the mapper for it. But at current state its ok for me.

We definitely need to create unit test for the new class and check if its correctly set in the ExampleControllerTest publish and unpublish action.

luca-rath:
> @nnatter @luca-rath Not sure if I would relly on it that the ContentCopier will use the persister and so the mapper for it. But at current state its ok for me.
> 
> We definitely need to create unit test for the new class and check if its correctly set in the ExampleControllerTest publish and unpublish action.

yes, of course. wanted to wait until writing the tests until the pr is approved.

nnatter:
> Not sure if I would relly on it that the ContentCopier will use the persister and so the mapper for it. But at current state its ok for me.

You are right. But I think the way this is implemented now, there is not hard dependency that the `ContentCopier` uses the `ContentPersister`. 

For example, if the `ContentCopier` would copy the raw data of the given `dimensionContentCollection`, the `published` value would still be copied. We would not even need the `WorkflowDataMapper` then 🙂

PR #140, opened 2020-05-08
Stop using preview cache

luca-rath:
After unserializing a serialized `ContentProjection` from preview cache, the embedded proxies e.g. of `excerptCategories` can't be lazyloaded anymore.


wachterjohannes:
@alexander-schranz @luca-rath will merge this as we should give it a try - for PHPCR preview implementation we introduced the cache to keep the database queries to phpcr as small as possible. with doctrine i dont think we have performance issues with this

alexander-schranz:
@wachterjohannes is ok for me :)

PR #144, opened 2020-05-15
Set workflow place only on draft dimension

### LINKED ISSUE #92: Set the current workflowPlace of the ContentWorkflow only on a specific DimensionContent

nnatter:
What do you think about this solution? This would fix #92

### DIFF
--- a/Content/Application/ContentDataMapper/DataMapper/WorkflowDataMapper.php
+++ b/Content/Application/ContentDataMapper/DataMapper/WorkflowDataMapper.php
@@ -86,7 +86,7 @@ class WorkflowDataMapper implements DataMapperInterface
         $published = $data['published'] ?? null;
 
         if (!$published) {
-            return;
+            throw new \RuntimeException('Expected "published" to be set in the data array.');
         }
 
         $object->setWorkflowPublished(new \DateTimeImmutable($published));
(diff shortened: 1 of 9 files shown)


alexander-schranz:
@nnatter is okay for me :)

nnatter:
added the missing test cases 🙂

PR #152, opened 2020-06-02
Fix getMediaId return type

Prokyonn:


Tests/Unit/Content/Infrastructure/Sulu/SmartContent/DataItem/ContentDataItemTest.php
```diff
@@ -95,7 +95,10 @@ public function testGetPublished(): void
 
         $dataItem = $this->getContentDataItem($contentProjection->reveal(), []);
 
-        $this->assertSame($published, $dataItem->getPublished());
+        /** @var \DateTimeImmutable $publishedDate */
+        $publishedDate = $dataItem->getPublished();
```
nnatter, line 99: would use `/* @phpstan-ignore-next-line */` like in the master branch. that way we will not forget about it when this is fixed in sulu 🙂 

https://github.com/sulu/SuluContentBundle/blob/3cc2a4c3bb865a854352db63bba6fa11c7f037e5/Tests/Unit/Content/Infrastructure/Sulu/SmartContent/DataItem/ContentDataItemTest.php#L116

PR #154, opened 2020-06-03
Add configurable route-schema

Prokyonn:
Adds the possibility to configure custom routes for the content entity.

Issue #113, opened 2020-03-08
NormalizeEnhancer should define the attributes to normalize


alexander-schranz:
Currently the NormalizeEnhancer only define which getters should not be called for normalization. Maybe the other way around would be better instead of defining what should be excluded it should be defined what should be included like we normally do it in sulu when using jms exlude all strategy. 

There are currently 2 way to define which attributes could be serialized. The first one is by using the **Normalizer Attributes** parameter `[AbstractNormalizer::ATTRIBUTES => ['template', 'templateData', ...]`.

The second one is using **Serialization Groups** introducing a `Content` serialization group would have the advantage that the developer could just use annotations and don't need to write a custom NormalizeEnhancer for just adding new property to the normalized data.

nnatter:
in general, i think things would be clearer if we change the behaviour to explicitly define attributes that should be normalized. right now, it feels kind of magic if you dont know that the symfony serializer is used 🙂 

but we should definitely avoid that the user needs to implement a `NormalizeEnhancer` for simple use cases.

niklasnatter:
A way for implementing this without the need of a `Normalizer` service would be adding a static `DimensionContentInterface::getNormalizeAttributes` method. 

The `ContentNormalizer` service then could use this method (and a similar method on all `Normalizer` services) to gather the attributes that should be included in the normalized data array.

niklasnatter:
After thinking about this once again, I concluded that it would probably be a cleaner solution to not use the `symfony/serializer` for normalization at all. I would propose to refactor the `NormalizerInterface` like this:

```diff
interface NormalizerInterface
{
-    /**
-     * @return string[]
-     */
-    public function getIgnoredAttributes(object $object): array;
-
-    /**
-     * @param mixed[] $normalizedData
-     *
-     * @return mixed[]
-     */
-    public function enhance(object $object, array $normalizedData): array;
-
+    /**
+     * @return mixed[]
+     */
+    public function normalize(object $object): array;
}
```

In additional, we can think about adding a `DimensionContentInterface::getAdditionalNormalizedData` method that allows to add normalized data to the result of the `ContentNormalizer`.

With this solution, the `ContentNormalizer` would just iterate over all `NormalizerInterface` services sorted by their priority and merge the data returned by the `NormalizerInterface::normalize` method.

I see the following advantages of such a solution:
- Simplifies `NormalizerInterface` and makes it consistent to the other interfaces
- No unexpected magic during normalization when adding a new property
[...]

PR #155, opened 2020-06-04
Add routable prefix to methods of RoutableInterface

nnatter:
I think we should prefix the methods (and therefore the keys of the normalized array) for all behaviours included in the bundle. The `RoutableInterface` is the only interface that did not use a prefix for its methods.

I see the following advantages:
* This prevents naming conflicts (method names and also keys in the normalized array)
* The `getIgnoredAttributes` method of the `Normalizer` classs are specific. At the moment, a project cannot normaliza a `contentId` because the `RoutableNormalizer` prevents it

### DIFF
--- a/Content/Domain/Model/RoutableInterface.php
+++ b/Content/Domain/Model/RoutableInterface.php
@@ -18,12 +18,12 @@ namespace Sulu\Bundle\ContentBundle\Content\Domain\Model;
  */
 interface RoutableInterface
 {
-    public function getRoutableResourceKey(): string;
+    public function getResourceKey(): string;
 
     /**
      * @return mixed
      */
     public function getRoutableId();
 
-    public function getRoutableLocale(): ?string;
+    public function getLocale(): ?string;
 }
--- a/Content/Domain/Model/RoutableInterface.php
+++ b/Content/Domain/Model/RoutableInterface.php
@@ -18,12 +18,12 @@ namespace Sulu\Bundle\ContentBundle\Content\Domain\Model;
  */
 interface RoutableInterface
 {
-    public function getResourceKey(): string;
+    public function getRoutableResourceKey(): string;
 
     /**
      * @return mixed
      */
-    public function getContentId();
+    public function getRoutableId();
 
-    public function getLocale(): ?string;
+    public function getRoutableLocale(): ?string;
 }
--- a/Tests/Application/ExampleTestBundle/Entity/ExampleDimensionContent.php
+++ b/Tests/Application/ExampleTestBundle/Entity/ExampleDimensionContent.php
@@ -100,9 +100,4 @@ class ExampleDimensionContent implements DimensionContentInterface, ExcerptInter
     {
         return Example::TEMPLATE_TYPE;
     }
-
-    public static function getContentClass(): string
-    {
-        return Example::class;
-    }
 }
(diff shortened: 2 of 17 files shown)


Content/Domain/Model/RoutableInterface.php
```diff
@@ -18,12 +18,12 @@
  */
 interface RoutableInterface
 {
-    public function getResourceKey(): string;
+    public function getRoutableResourceKey(): string;
```
nnatter, line 21: i am thinking about moving the `getResourceKey` method from the `ContentRichEntityInterface` to the `DimensionContentInterface` in another pull request. The `getResourceKey` method shoukd be on the same entity as the `getTemplateType` and `getWorkflowName` IMO. what do you think?

Tests/Functional/Integration/responses/example_get.json
```diff
@@ -14,7 +14,6 @@
     "locale": "en",
     "published": null,
     "publishedState": false,
-    "resourceKey": "@string@",
```
nnatter, line 17: This was added with #154 which was merged just an hour ago. So i think it should not be a big BC break 🙂

Tests/Application/ExampleTestBundle/Entity/ExampleDimensionContent.php
```diff
@@ -100,9 +100,4 @@ public static function getTemplateType(): string
     {
         return Example::TEMPLATE_TYPE;
     }
-
-    public static function getContentClass(): string
```
nnatter, line 104: we forgot about this in #154

Tests/Unit/Mocks/ContentRichEntityMockWrapperTrait.php
```diff
@@ -0,0 +1,62 @@
...
+use Doctrine\Common\Collections\Collection;
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\DimensionContentInterface;
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\DimensionInterface;
+
+/**
+ * Trait for composing a class that wraps a ContentRichEntityInterface mock.
+ *
+ * @see MockWrapper to learn why this trait is needed.
+ */
+trait ContentRichEntityMockWrapperTrait
```
nnatter, line 25: this wozld not be necessary if we move the `getResourceKey` method

nnatter:
Not sure about that - if two behaviour interfaces (eg `TemplateInterface` and `RoutableInterface`) have a `getResourceKey` method, we cannot add the `getResourceKey` method to both traits because they would conflict. Maybe these unprefixed methods (resource key, locale, ...) should be on the `DimensionContentInterface`?

But i think that is a discussion for another day 🙂 I adjusted this PR to only rename the `getContentId` to `getRoutableId`. Would like to get rid of the `contentId` term because we dont use that anywhere else and it might be confusing..

alexander-schranz:
The issue we have with serialization is what we have documented #113 that we should define what should be serialized and not what should not be serialized. Not 100% happy with `getRoutableId` but we could go with that currently.

nnatter:
Do you have a better idea than `getRoutableId`? 🙂

alexander-schranz:
@nnatter currently not 🙈

PR #156, opened 2020-06-05
Enhance route generation

luca-rath:


Content/Application/ContentDataMapper/DataMapper/RoutableDataMapper.php
```diff
@@ -138,8 +146,14 @@ public function map(
 
         $routePath = $data[$name] ?? null;
         if (!$routePath) {
+            /** @var mixed $routeGenerationData */
+            $routeGenerationData = array_merge(
```
nnatter, line 150: unfortunately this will discard most of the unlocalized data, especially unlocalized template properties. in theory, we should use our mergers here, but we would need to create a `DimensionContentCollectionInterface` which is kind of strange too 😕

Content/Application/ContentDataMapper/DataMapper/RoutableDataMapper.php
```diff
@@ -138,8 +146,14 @@ public function map(
...
+                $this->contentNormalizer->normalize($unlocalizedObject),
+                $this->contentNormalizer->normalize($localizedObject)
+            );
+
             $routePath = $this->routeGenerator->generate(
-                $localizedObject,
+                $routeGenerationData,
```
nnatter, line 156: i think passing somethink like this here would be more pragmatic:
```
array_merge(
    $data,
    ['unlocalizedDimensionContent' => $unlocalizedObject, 'localizedDimensionContent' => $localizedObject]
)
```

PR #160, opened 2020-06-22
Add cache-lifetime-resolve to route defaults provider

wachterjohannes:


Content/Infrastructure/Sulu/Route/ContentRouteDefaultsProvider.php
```diff
@@ -84,6 +93,7 @@ public function getByEntity($entityClass, $id, $locale, $object = null)
             'view' => $structureBridge->getView(),
             'structure' => $structureBridge,
             '_controller' => $structureBridge->getController(),
+            '_cacheLifetime' => $this->getCacheLifetime($structureBridge->getStructure()),
```
nnatter, line 96: would check this value in the test too

Content/Infrastructure/Sulu/Structure/ContentStructureBridge.php
```diff
@@ -77,6 +77,11 @@ public function getDocument(): ContentDocument
         return new ContentDocument($this->content, $this->locale);
     }
 
+    public function getStructure(): StructureMetadata
```
nnatter, line 80: not sure if we have tested this class. if yes, we should add a unit test for this too 🙂

PR #161, opened 2020-07-02
Add symfony 5 compatibility

luca-rath:


luca-rath:
@alexander-schranz Pipeline fails because of https://github.com/handcraftedinthealps/code-coverage-checker ... Can you check that?

alexander-schranz:
@luca-rath Thank you!

PR #163, opened 2020-07-23
Rename getRoutableId method of RoutableInterface to getResourceId

nnatter:
This makes the naming consistent with the `getResourceKey` method of the `RoutableInterface` and the `DimensionContentInterface`.


alexander-schranz:
Like 👍 this as maybe other interfaces needing the `getResoureId` can require the same method so no duplicated methods needed.

Issue #167, opened 2020-09-16
Resolving of DimensionContent class can fail when using inheritance


mgwestwerk:
In cases where the doctrine inheritance feature is used the `targetEntity` of the `dimensionContents`-Relation can be an abstract class which does not implement all methods of the `DimensionContentInterface`.
This can lead to errors for example when trying to get the resource key for a given `ContentRichEntity`

Possible Solution:
Move the functionality to determine the associated dimension content class to a single locaition (e.g. a service) so that it can be overriden/extended

nnatter:
Its possible to override the `ContentMetadataInspector` service and return the correct metadata now as a workaround (#168). Still, it would be nice to properly support single table inheritance mappings.

PR #168, opened 2020-09-16
Add ContentMetadataInspector service to provide correct target classes

mgwestwerk:
#167

### DIFF
--- a/Content/Application/ContentMetadataInspector/ContentMetadataInspector.php
+++ b/Content/Application/ContentMetadataInspector/ContentMetadataInspector.php
@@ -27,11 +27,6 @@ class ContentMetadataInspector implements ContentMetadataInspectorInterface
         $this->entityManager = $entityManager;
     }
 
-    /**
-     * @param class-string<ContentRichEntityInterface> $contentRichEntityClass
-     *
-     * @return class-string<DimensionContentInterface>
-     */
     public function getDimensionContentClass(string $contentRichEntityClass): string
     {
         $classMetadata = $this->entityManager->getClassMetadata($contentRichEntityClass);
@@ -40,9 +35,6 @@ class ContentMetadataInspector implements ContentMetadataInspectorInterface
         return $associationMapping['targetEntity'];
     }
 
-    /**
-     * @param class-string<ContentRichEntityInterface> $contentRichEntityClass
-     */
     public function getDimensionContentMappingProperty(string $contentRichEntityClass): string
     {
         $classMetadata = $this->entityManager->getClassMetadata($contentRichEntityClass);
--- a/Content/Application/ContentMetadataInspector/ContentMetadataInspector.php
+++ b/Content/Application/ContentMetadataInspector/ContentMetadataInspector.php
@@ -22,10 +22,11 @@ class ContentMetadataInspector implements ContentMetadataInspectorInterface
      */
     private $entityManager;
 
-    function __construct(EntityManagerInterface $entityManager) {
+    public function __construct(EntityManagerInterface $entityManager)
+    {
         $this->entityManager = $entityManager;
     }
-    
+
     /**
      * @param class-string<ContentRichEntityInterface> $contentRichEntityClass
      *
@@ -38,11 +39,9 @@ class ContentMetadataInspector implements ContentMetadataInspectorInterface
 
         return $associationMapping['targetEntity'];
     }
-    
+
     /**
      * @param class-string<ContentRichEntityInterface> $contentRichEntityClass
-     *
-     * @return string
      */
     public function getDimensionContentMappingProperty(string $contentRichEntityClass): string
     {
--- a/Content/Application/ContentMetadataInspector/ContentMetadataInspector.php
+++ b/Content/Application/ContentMetadataInspector/ContentMetadataInspector.php
@@ -11,11 +11,11 @@ declare(strict_types=1);
  * with this source code in the file LICENSE.
  */
 
-namespace Sulu\Bundle\ContentBundle\Content\Application\ContentAssociationMapper;
+namespace Sulu\Bundle\ContentBundle\Content\Application\ContentMetadataInspector;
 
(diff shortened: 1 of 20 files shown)


Content/Application/ContentMetadataInspector/ContentMetadataInspectorInterface.php
```diff
@@ -0,0 +1,32 @@
...
+     * @param class-string<ContentRichEntityInterface> $contentRichEntityClass
+     *
+     * @return class-string<DimensionContentInterface>
+     */
+    public function getDimensionContentClass(string $contentRichEntityClass): string;
+
+    /**
+     * @param class-string<ContentRichEntityInterface> $contentRichEntityClass
+     */
+    public function getDimensionContentMappingProperty(string $contentRichEntityClass): string;
```
nnatter, line 31: think i would call this `getDimensionContentPropertyName`

Tests/Unit/Content/Application/ContentMetadataInspector/ContentMetadataInspectorTest.php
```diff
@@ -0,0 +1,51 @@
...
+{
+    protected function createContentMetadataInspectorTestInstance(
+        EntityManagerInterface $entityManager
+    ): ContentMetadataInspectorInterface {
+        return new ContentMetadataInspector(
+            $entityManager
+        );
+    }
+
+    public function testGetDimensionContentClass(): void
```
nnatter, line 34: test case for the second method is missing

nnatter (commented):
looks good to me otherwise 🙂  /cc @alexander-schranz

nnatter:
thanks for working on this! @mgwestwerk

Content/Application/ContentMetadataInspector/ContentMetadataInspector.php
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\ContentBundle\Content\Application\ContentMetadataInspector;
+
+use Doctrine\ORM\EntityManagerInterface;
+
+class ContentMetadataInspector implements ContentMetadataInspectorInterface
```
alexander-schranz, line 18:
```suggestion

/**
 * @internal This class is for internal use and we can not given any bc promise for this as it can change.
 */
class ContentMetadataInspector implements ContentMetadataInspectorInterface
```

Content/Application/ContentMetadataInspector/ContentMetadataInspectorInterface.php
```diff
@@ -0,0 +1,32 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\ContentBundle\Content\Application\ContentMetadataInspector;
+
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\ContentRichEntityInterface;
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\DimensionContentInterface;
+
+interface ContentMetadataInspectorInterface
```
alexander-schranz, line 19:
```suggestion

/**
 * @internal This interface is for internal use and we can not given any bc promise for this as it can change.
 */
interface ContentMetadataInspectorInterface
```

Content/Application/ContentMetadataInspector/ContentMetadataInspectorInterface.php
alexander-schranz: agree we should be as specific as possible

Resources/config/services.xml
```diff
@@ -57,9 +57,17 @@
 
         <service id="Sulu\Bundle\ContentBundle\Content\Application\ContentDataMapper\ContentDataMapperInterface" alias="sulu_content.content_data_mapper"/>
 
+        <!-- Content Metadata Inspector -->
+        <service id="sulu_content.content_metadata_inspector" class="Sulu\Bundle\ContentBundle\Content\Application\ContentMetadataInspector\ContentMetadataInspector">
+            <argument type="service" id="doctrine.orm.entity_manager"/>
+        </service>
+
+        <service id="Sulu\Bundle\ContentBundle\Content\Application\ContentMetadataInspector\ContentMetadataInspectorInterface" alias="sulu_content.content_metadata_inspector"/>
```
alexander-schranz, line 65: I would not make this service autowireable so this should not exist.

Resources/config/services.xml
```diff
@@ -57,9 +57,17 @@
 
         <service id="Sulu\Bundle\ContentBundle\Content\Application\ContentDataMapper\ContentDataMapperInterface" alias="sulu_content.content_data_mapper"/>
 
+        <!-- Content Metadata Inspector -->
+        <service id="sulu_content.content_metadata_inspector" class="Sulu\Bundle\ContentBundle\Content\Application\ContentMetadataInspector\ContentMetadataInspector">
```
alexander-schranz, line 61: I'm not sure if I would make a service for it as its internal logic, I would make construct it where needed in the `__construct` function of that service.

@nnatter what do you think?

alexander-schranz (changes requested):
@mgwestwerk Thank you for working for this. Some small change requests and some things to discuss.

Resources/config/services.xml
nnatter: that way it is not possible to overwrite the service

mgwestwerk: To give you my reasoning, why i made it a service: My problem that i described in #168 requires me to override the default functionality that maps the `ContentRichEntity` class to the `DimensionContent` class. By putting it into a service I was able to decorate it and achieve this goal.

alexander-schranz:
I created an issue for it (#170). At current state its okay for me but hope we get ride of it soon and detect the target classes correctly then.

Issue #145, opened 2020-05-15
Throw Exception if there is no DimensionContent for the given dimensionAttributes


nnatter:
The `ContentResolver` service allows to resolve the **effective content** for a given `ContentRichEntity` and given `dimensionAttributes`. In this context, **effective content** means that the object returned by the `ContentResolver` is a merged representation/projection of all relevant `DimensionContent`s (eg. localized `DimensionContent` and unlocalized `DimensionContent`) for the given `dimensionAttributes`.

The `ContentResolver` throws an `ContentNotFoundException`, when there are no relevant `DimensionContent`s found. At the moment, **no exception is thrown** if there is at least one relevant `DimensionContent`. 

This means that the `ContentResolver` does not throw an exception when a developer queries for the published german content with the dimensionAttributes `['stage' => 'live', 'locale' => 'de']` if there is at least one published language (even if the german content is not published). Because of this, we need to explicitly check the resolved content with the following code in various places of the bundle and also in our projects:

```php
/** @var ThingProjection $thingProjection */
$thingProjection = $this->contentManager->resolve($thing, [
    'locale' => $locale,
    'stage' => DimensionInterface::STAGE_LIVE,
]);

$dimension = $thingProjection->getDimension();

if (DimensionInterface::STAGE_LIVE !== $dimension->getStage() || $locale !== $dimension->getLocale()) {
    return null;
}

return $thingProjection;
```

In my opinion, this is not not a great DX. Firstly, as a developer, I do not expect this behaviour and therefore it is really easy to forget about the check. Secondly, the consequences of forgetting about the check are quite big. 

As there are still places where the current behaviour is desired, a way for solving this would be adding a `throwIfDimensionAttributesDoNotMatch` flag (hopefully we will find a better name) to the `resolve()` method of the `ContentResolver`. This would lead to the following method signature:

```php
public function resolve(ContentRichEntityInterface $contentRichEntity, array $dimensionAttributes, bool $throwIfDimensionAttributesDoNotMatch = true): ContentProjectionInterface;
```

alexander-schranz:
For me this is a little bit related to the options parameters #136.
The document manager provides also options e.g.: `['load_ghost_content' => false]`. I think something similar does make sense.

```php
public function resolve(
    ContentRichEntityInterface $contentRichEntity, 
    array $dimensionAttributes, 
    array $options = []
): ContentProjectionInterface;
```

PR #173, opened 2020-12-10
Improve tests and fix code coverage

alexander-schranz:


Content/Infrastructure/Sulu/SmartContent/Repository/ContentDataProviderRepository.php
```diff
@@ -113,7 +113,7 @@ function (ContentRichEntityInterface $contentRichEntity) use ($locale, $showUnpu
                     $dimension = $resolvedDimensionContent->getDimension();
 
                     if ($stage !== $dimension->getStage() || $locale !== $dimension->getLocale()) {
-                        return null;
+                        return null; // @codeCoverageIgnore
```
alexander-schranz, line 116: I'm not sure if this if statement is needed we query for `stage` and the correct `locale` in the queryBuilder. When here the false stage and locale would be returned it would be a bug in the `contentResolver` and should throw an exception but I don't see any need for exception or check here. 

@nnatter would you remove the if else statement? /cc @luca-rath any knowledge why it was added?

luca-rath: I'm not sure if it's still the case, but at some point it was possible to query for `['locale' => 'en', 'stage' => 'live']` and the result was a `DimensionContent` with these `['locale' => null, 'stage' => 'live']` dimension attributes ... This happened, if the entity was not published in the requested locale but was published in any other locale

alexander-schranz: Thx for feedback this should not happen because of the following where statement: https://github.com/sulu/SuluContentBundle/blob/0e9ebd59156c19b5469f3706fb89b8af0216e7b9/Content/Infrastructure/Sulu/SmartContent/Repository/ContentDataProviderRepository.php#L453

If this would really be the case we would need to have a look again at that query, or the contentResolver / merger.

nnatter: I think it is still the case. See https://github.com/sulu/SuluContentBundle/issues/145

alexander-schranz: @nnatter the query itself should not return any `ids` which has that so #145 should not happen here.

nnatter: Okay thats possible - have not looked into the implementation of the `ContentDataProviderRepository` 🙂

luca-rath: I guess it would make sense to introduce a new argument like $allowPartial or something like that for the `ContentResolver::resolve()`, so the ContentResolver can make that above check for you and throw an exception if it fails.
Because sometimes it would totally make sense, if I get only a partial merged DimensionContent from the ContentResolver. An example would be, if I switch to a new locale in the admin, the unlocalized fields should already be prefilled, even if that entity doesn't exist in that locale.

alexander-schranz: @luca-rath we will definitely add an options array as things like `load_ghost` and `load_shadow` need to be handled in future over the resolver. So I would not introduce here a specific boolean flag. But thats part of #145

Content/Infrastructure/Sulu/SmartContent/Repository/ContentDataProviderRepository.php
```diff
@@ -209,9 +214,8 @@ protected function findEntityIdsByFilters(
             );
         }
 
-        $query = $queryBuilder->getQuery();
         foreach ($parameters as $parameter => $value) {
-            $query->setParameter($parameter, $value);
+            $queryBuilder->setParameter($parameter, $value);
```
nnatter, line 218: never understood why i can set this on the `Query` and also the `QueryBuilder`. does this make a difference? 🙈

Content/Infrastructure/Sulu/Teaser/ContentTeaserProvider.php
```diff
@@ -56,21 +56,29 @@ abstract class ContentTeaserProvider implements TeaserProviderInterface
...
     protected $contentRichEntityClass;
 
+    /**
+     * @var bool
+     */
+    protected $showDrafts;
+
     /**
      * @param class-string<ContentRichEntityInterface> $contentRichEntityClass
+     * @param bool $showDrafts Inject parameter "sulu_document_manager.show_drafts" here
```
nnatter, line 66: 🆒

Content/Infrastructure/Sulu/SmartContent/Repository/ContentDataProviderRepository.php
alexander-schranz: Don't know I think it should always be done by the QueryBuilder not on the Query itself.

nnatter (approved):

PR #177, opened 2021-01-08
Refactor composer lint script and add php-cs-fixer as dev-dependency

niklasnatter:


Tests/Unit/Content/Application/ContentWorkflow/ContentWorkflowTest.php
```diff
@@ -55,7 +55,8 @@ protected function createContentWorkflowInstance(
      */
     protected function wrapWorkflowMock(ObjectProphecy $workflowMock)
     {
-        return new class($workflowMock) extends MockWrapper implements DimensionContentInterface,
+        return new class($workflowMock) extends MockWrapper implements
+    DimensionContentInterface,
```
alexander-schranz, line 59: This should be indented correctly the fixer currently always adds only 4 spaces instead of the needed one for anonoymous class.

nnatter: done 🙂

PR #180, opened 2021-01-15
Fix create tag with same name multiple times

### LINKED ISSUE #176: Duplicated Tag unique constraint error when creating multiple Content with same Tag
When we are creating a Document 1 with a Tag `A` and create a Document 2 also with a Tag `A` and the Tag `A` did not exist the TagFactory currently will create for Document 1 and Document 2 a new Tag aslong as between no "flush" of the entityManager was called and will error in a UniqueConstraint.

This can be fixed by #175 or a low version of it by just cache newly created tags and check for newly tags in the [TagFactory](https://github.com/sulu/SuluContentBundle/blob/4d62683790248bda95fa243ad34b59078c79a2a2/Content/Infrastructure/Doctrine/TagFactory.php#L59-L73):

```diff
        foreach ($tagNames as $key => $tagName) {
            if (isset($excerptTags[$key])) {
                continue;
            }
            
+            $tag = $this->createdTags[$tag->getName()] ?? null;

+            if (!$tag) {
                /** @var TagInterface $tag */
                $tag = $this->tagRepository->createNew();
                $tag->setName($tagName);
   
                $this->entityManager->persist($tag);
+               $this->createdTags[$tag->getName()] = $tag;
+            }

            $excerptTags[$key] = $tag;
        }
```

alexander-schranz:
Fix create tag with same name multiple times, without flushing anything.

Fixes #176

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -78,7 +78,8 @@
             "Tests/Application/bin/adminconsole sulu:build dev --env dev"
         ],
         "bootstrap-test-environment": [
-            "Tests/Application/bin/adminconsole doctrine:database:create --if-not-exists --env test",
+            "Tests/Application/bin/adminconsole doctrine:database:drop --if-exists --force --env test",
+            "Tests/Application/bin/adminconsole doctrine:database:create --env test",
             "Tests/Application/bin/adminconsole doctrine:schema:update --force --env test"
         ],
         "lint": [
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -78,7 +78,8 @@
             "Tests/Application/bin/adminconsole sulu:build dev --env dev"
         ],
         "bootstrap-test-environment": [
-            "Tests/Application/bin/adminconsole doctrine:database:create --if-not-exists --env test",
```
alexander-schranz, line 81: schema update not always work when having foreign key references so I drop in the bootstrap test command the whole database and recreate it.

niklasnatter (approved):

Issue #178, opened 2021-01-13
Simplify the DimensionContentCollection class


niklasnatter:
After https://github.com/sulu/SuluContentBundle/pull/174 is merged, the `DimensionContentCollection` class will be one of the main classes of the bundle. Because of this, I think it is important to keep the class simple and consistent. Therefore I would propose the following changes:

- [ ] a) The `DimensionContentCollection` should not extend `\Traversable` or `\Countable`. This makes the implementation of the class more complicated without improving DX. As a developer, I would expect a `getDimensionContents` method that returns an array of `DimensionContent`s.

- [x] b) Remove the `getUnlocalizedDimensionContent` method and `getLocalizedDimensionContent` method. To keep things flexible, all services should use the `getDimensionContent` method with a respective `dimensionAttributes` parameter (see https://github.com/sulu/SuluContentBundle/issues/84#issuecomment-759571798). By removing the methods, we force our services to consistently handle the `dimensionAttributes` parameter. (fixed in https://github.com/sulu/SuluContentBundle/pull/182)

After these changes, the `DimensionContentCollectionInterface` should look something like this:
```php
interface DimensionContentCollectionInterface
{
    /**
     * @return DimensionContentInterface[]
     */
    public function getDimensionContents(): array;

    /**
     * @param mixed[] $dimensionAttributes
     */
    public function getDimensionContent(array $dimensionAttributes): ?DimensionContentInterface;

    /**
     * @return class-string<DimensionContentInterface>
     */
    public function getDimensionContentClass(): string;

    /**
     * @return mixed[]
     */
    public function getDimensionAttributes(): array;
}
```

PR #182, opened 2021-01-18
Pass DimensionContentCollection to DataMapper services

niklasnatter:
This PR aadjust the `ContentDataMapper` service and the `DataMapper` services to accept a `DimensionContentCollection` parameter instead of a `unlocalizedObject` and an optional `localizedObject` parameter.

This makes the interfaces more flexible and consistent to other parts of the bundle.

### DIFF
--- a/Content/Application/ContentDataMapper/ContentDataMapperInterface.php
+++ b/Content/Application/ContentDataMapper/ContentDataMapperInterface.php
@@ -13,6 +13,8 @@ declare(strict_types=1);
 
 namespace Sulu\Bundle\ContentBundle\Content\Application\ContentDataMapper;
 
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\DimensionContentCollectionInterface;
+
 interface ContentDataMapperInterface
 {
     /**
@@ -20,7 +22,6 @@ interface ContentDataMapperInterface
      */
     public function map(
         array $data,
-        object $unlocalizedObject,
-        ?object $localizedObject = null
+        DimensionContentCollectionInterface $dimensionContentCollection
     ): void;
 }
--- a/Content/Application/ContentDataMapper/DataMapper/DataMapperInterface.php
+++ b/Content/Application/ContentDataMapper/DataMapper/DataMapperInterface.php
@@ -13,6 +13,8 @@ declare(strict_types=1);
 
 namespace Sulu\Bundle\ContentBundle\Content\Application\ContentDataMapper\DataMapper;
 
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\DimensionContentCollectionInterface;
+
 interface DataMapperInterface
 {
     /**
@@ -20,7 +22,6 @@ interface DataMapperInterface
      */
     public function map(
         array $data,
-        object $unlocalizedObject,
-        ?object $localizedObject = null
+        DimensionContentCollectionInterface $dimensionContentCollection
     ): void;
 }
--- a/Content/Domain/Model/DimensionContentCollection.php
+++ b/Content/Domain/Model/DimensionContentCollection.php
@@ -87,11 +87,6 @@ class DimensionContentCollection implements \IteratorAggregate, DimensionContent
         return $this->dimensionContentClass;
     }
 
-    public function getLocalizedDimensionContent(): ?DimensionContentInterface
-    {
-        return $this->localizedDimensionContent;
-    }
-
     public function getDimensionContent(array $dimensionAttributes): ?DimensionContentInterface
     {
         $dimensionAttributes = array_merge($this->defaultDimensionAttributes, $dimensionAttributes);
--- a/Content/Domain/Model/DimensionContentCollection.php
+++ b/Content/Domain/Model/DimensionContentCollection.php
@@ -87,11 +87,6 @@ class DimensionContentCollection implements \IteratorAggregate, DimensionContent
         return $this->dimensionContentClass;
     }
 
-    public function getUnlocalizedDimensionContent(): ?DimensionContentInterface
-    {
-        return $this->unlocalizedDimensionContent;
-    }
-
     public function getLocalizedDimensionContent(): ?DimensionContentInterface
     {
         return $this->localizedDimensionContent;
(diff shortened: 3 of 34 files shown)


Content/Infrastructure/Sulu/Preview/PreviewDimensionContentCollection.php
```diff
@@ -0,0 +1,67 @@
...
+namespace Sulu\Bundle\ContentBundle\Content\Infrastructure\Sulu\Preview;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\DimensionContentCollectionInterface;
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\DimensionContentInterface;
+
+/**
+ * @implements \IteratorAggregate<DimensionContentInterface>
+ */
+class PreviewDimensionContentCollection implements \IteratorAggregate, DimensionContentCollectionInterface
```
alexander-schranz, line 24: Not sure about this class. Can we mark it as `@internal` at current state.

Content/Infrastructure/Sulu/Preview/PreviewDimensionContentCollection.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public function getDimensionContent(array $dimensionAttributes): ?DimensionContentInterface
+    {
+        return $this->previewDimensionContent;
+    }
+
+    public function getDimensionAttributes(): array
+    {
+        return ['locale' => $this->previewLocale];
```
alexander-schranz: A DimensionCollection should always return all its `DimensionAttributes` so we should `array_merge()` them here with the `getDefaultDimensionAttributes`.

Content/Infrastructure/Sulu/Preview/PreviewDimensionContentCollection.php
```diff
@@ -0,0 +1,67 @@
...
+
+    public function getIterator()
+    {
+        return new ArrayCollection([$this->previewDimensionContent]);
+    }
+
+    public function count(): int
+    {
+        return 1;
+    }
```
alexander-schranz, line 70: Not really happy with this :(

alexander-schranz: Not sure if we should return here `2` to simulate that we have a unlocalized and localizedObject.

alexander-schranz (changes requested):
I personally not so happy with the Preview Handling, but it was also not good before. I would atleast simulate the count: 2 and mark the PreviewCollection as internal so we can change it the way we need it in future.

niklasnatter:
@alexander-schranz Not really happy about the preview handling too, but I dont have a better idea at the moment. Do you see a better alternative?

Content/Infrastructure/Sulu/Preview/PreviewDimensionContentCollection.php
niklasnatter: 👍

Content/Infrastructure/Sulu/Preview/PreviewDimensionContentCollection.php
niklasnatter: 👍

Content/Infrastructure/Sulu/Preview/PreviewDimensionContentCollection.php
niklasnatter: think i would not return 2 because that is misleading

Content/Application/ContentDataMapper/ContentDataMapperInterface.php
```diff
@@ -13,14 +13,15 @@
...
 namespace Sulu\Bundle\ContentBundle\Content\Application\ContentDataMapper;
 
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\DimensionContentCollectionInterface;
+
 interface ContentDataMapperInterface
 {
     /**
      * @param array<string, mixed> $data
      */
     public function map(
```
niklasnatter, line 23: TODO: document this in the `UPGRADE.md`

Content/Application/ContentDataMapper/DataMapper/DataMapperInterface.php
```diff
@@ -13,14 +13,15 @@
...
 namespace Sulu\Bundle\ContentBundle\Content\Application\ContentDataMapper\DataMapper;
 
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\DimensionContentCollectionInterface;
+
 interface DataMapperInterface
 {
     /**
      * @param array<string, mixed> $data
      */
     public function map(
```
niklasnatter, line 23: TODO: document this in the `UPGRADE.md`

Content/Domain/Model/DimensionContentCollection.php
```diff
@@ -87,16 +87,6 @@ public function getDimensionContentClass(): string
         return $this->dimensionContentClass;
     }
 
-    public function getUnlocalizedDimensionContent(): ?DimensionContentInterface
```
nnatter, line 90: TODO: document this in the `UPGRADE.md`

Content/Domain/Model/DimensionContentCollectionInterface.php
```diff
@@ -18,10 +18,6 @@
  */
 interface DimensionContentCollectionInterface extends \Traversable, \Countable
 {
-    public function getUnlocalizedDimensionContent(): ?DimensionContentInterface;
```
niklasnatter, line 21: TODO: document this in the `UPGRADE.md`

Content/Infrastructure/Sulu/Preview/PreviewDimensionContentCollection.php
niklasnatter: i could adjust the `DimensionContentCollectionInterface` to not extend `Countable` as described in https://github.com/sulu/SuluContentBundle/issues/178 to get rid of the method 😇

PR #187, opened 2021-02-12
Add PHP 8 compatibility

alexander-schranz:
- [x] Add Test Matrix (Github Actions)


.github/workflows/test-application.yaml
```diff
@@ -0,0 +1,108 @@
...
+              uses: ramsey/composer-install@v1
+              with:
+                  dependency-versions: ${{ matrix.dependency-versions }}
+                  composer-options: ${{ matrix.composer-options }}
+
+            - name: Bootstrap test environment
+              run: composer bootstrap-test-environment
+
+            - name: Execute test
+              if: matrix.coverage == false
```
alexander-schranz, line 80: any hint to check for false here? tried different things with and without ${{ }} but nothing did work /cc @luca-rath

alexander-schranz: okay did now work this way 🙈

.coveralls.yml
```diff
@@ -0,0 +1,3 @@
+service_name: github-actions
```
alexander-schranz: Copied from: https://github.com/stripe/stripe-php/blob/master/.coveralls.github-actions.yml

.github/workflows/test-application.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+            - name: Execute test with coverage
+              if: ${{ matrix.coverage }}
+              run: composer test-with-coverage
+
+            - name: Lint Code
+              if: ${{ matrix.lint }}
+              run: composer lint
+
+            - name: Coverage all upload
```
alexander-schranz: See https://github.com/php-coveralls/php-coveralls#github-actions

.github/workflows/test-application.yaml
```diff
@@ -0,0 +1,100 @@
...
+
+            - name: Execute test
+              if: matrix.coverage == false
+              run: composer test
+
+            - name: Execute test with coverage
+              if: ${{ matrix.coverage }}
+              run: composer test-with-coverage
+
+            - name: Lint Code
```
niklasnatter: think we executed the lint script before the test script in other repositories

.github/workflows/test-application.yaml
```diff
@@ -0,0 +1,100 @@
...
+            - name: Install composer dependencies
+              uses: ramsey/composer-install@v1
+              with:
+                  dependency-versions: ${{ matrix.dependency-versions }}
+                  composer-options: ${{ matrix.composer-options }}
+
+            - name: Bootstrap test environment
+              run: composer bootstrap-test-environment
+
+            - name: Execute test
```
niklasnatter:
```suggestion
            - name: Execute test cases
```

.github/workflows/test-application.yaml
```diff
@@ -0,0 +1,100 @@
...
+              if: matrix.coverage == false
+              run: composer test
+
+            - name: Execute test with coverage
```
niklasnatter:
```suggestion
            - name: Execute test cases with coverage
```

.github/workflows/test-application.yaml
```diff
@@ -0,0 +1,100 @@
...
+
+            - name: Coverage all upload
+              if: ${{ matrix.coverage }}
+              run: |
+                  composer global require php-coveralls/php-coveralls
+                  php-coveralls -vvv
+              env:
+                  COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+            - name: Check coverage
```
niklasnatter:
```suggestion
            - name: Check code coverage
```

.github/workflows/test-application.yaml
```diff
@@ -0,0 +1,100 @@
...
+
+            - name: Execute test with coverage
+              if: ${{ matrix.coverage }}
+              run: composer test-with-coverage
+
+            - name: Lint Code
+              if: ${{ matrix.lint }}
+              run: composer lint
+
+            - name: Coverage all upload
```
niklasnatter:
```suggestion
            - name: Upload code coverage to coveralls
```

.github/workflows/test-application.yaml
```diff
@@ -0,0 +1,100 @@
...
+              if: ${{ matrix.coverage }}
+              run: |
```
niklasnatter, line 92: think i would use the official Coveralls github action for uploading 😇 https://github.com/marketplace/actions/coveralls-github-action

.github/workflows/test-application.yaml
alexander-schranz: did move it after the tests so the tests are always executed and linting of code style is done after it 🤔

.github/workflows/test-application.yaml
alexander-schranz: @nnatter that doesn't support PHPUNIT xml reports

alexander-schranz: it supports just the `lcov` / `./coverage/lcov.info` format. So basically not PHP support which can only xml, json, html out of the box..

Open Issue: https://github.com/coverallsapp/github-action/issues/15

.github/workflows/test-application.yaml
alexander-schranz: anyway moved it up ;)

.coveralls.yml
```diff
@@ -0,0 +1,3 @@
...
+coverage_clover: Tests/reports/clover.xml
```
niklasnatter: think we could add these as argument to the `php-coveralls` call. but is okay for me this way too
https://github.com/coverallsapp/github-action/issues/15#issuecomment-643747251

.github/workflows/test-application.yaml
niklasnatter: you are right - sorry did not see that

niklasnatter (approved):

.coveralls.yml
alexander-schranz: I will try it if it works not sure the `service_name` can not be set and maybe thats required in this case that the PR is correctly updated, lets have a look if it works or not without the config.

PR #188, opened 2021-02-15
Add event-subscriber to throw flushed workflow events

### LINKED ISSUE #129: Add post flush event for transitions
Sometimes you need to be sure that the content entity is flushed to the database when a transition did happened and then do something specific e.g. build a search index. For this we should a doctrine flush subscriber which listening to transition events and throw a new workflow transition post flush (not sure about the naming) event.

wachterjohannes:
Fixes #129

### DIFF
--- a/Content/Infrastructure/Symfony/Workflow/FlushedEventSubscriber.php
+++ b/Content/Infrastructure/Symfony/Workflow/FlushedEventSubscriber.php
@@ -1,0 +1,76 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\ContentBundle\Content\Infrastructure\Symfony\Workflow;
+
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\WorkflowInterface;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Workflow\Event\Event;
+
+/**
+ * @internal
+ */
+class FlushedEventSubscriber implements EventSubscriberInterface
+{
+    /**
+     * @return string[]
+     */
+    public static function getSubscribedEvents()
+    {
+        return [
+            'workflow.completed' => 'onCompleted',
+        ];
+    }
+
+    /**
+     * @var EventDispatcherInterface
+     */
+    private $eventDispatcher;
+
+    /**
+     * @var mixed[]
+     */
+    private $eventsToDispatch = [];
+
+    public function __construct(EventDispatcherInterface $eventDispatcher)
+    {
+        $this->eventDispatcher = $eventDispatcher;
+    }
+
+    public function onCompleted(Event $event): void
+    {
+        if (WorkflowInterface::WORKFLOW_DEFAULT_NAME !== $event->getWorkflowName()) {
+            return;
+        }
+
+        $eventName = sprintf('workflow.%s.flushed', $event->getWorkflowName());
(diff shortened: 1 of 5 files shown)


Content/Infrastructure/Symfony/Workflow/FlushedEventSubscriber.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace Sulu\Bundle\ContentBundle\Content\Infrastructure\Symfony\Workflow;
+
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\WorkflowInterface;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Workflow\Event\Event;
+
+class FlushedEventSubscriber implements EventSubscriberInterface
```
alexander-schranz, line 24: Mark this service as `@internal`. This is no class somebody should require or extend from. Its just a internal service which should throw that specific events.

Content/Infrastructure/Symfony/Workflow/FlushedEventSubscriber.php
```diff
@@ -0,0 +1,53 @@
...
+{
+    /**
+     * @return string[]
+     */
+    public static function getSubscribedEvents()
+    {
+        return [
+            sprintf('workflow.%s.completed', WorkflowInterface::WORKFLOW_DEFAULT_NAME) => 'onCompleted',
```
alexander-schranz: create an issue that this is here hardcoded and we should find a way that this events are also thrown for custom workflows

Content/Infrastructure/Symfony/Workflow/FlushedEventSubscriber.php
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public function onFlush(): void
+    {
+        foreach ($this->eventsToDispatch as $item) {
+            $this->eventDispatcher->dispatch($item[0], $item[1]);
+        }
+
+        $this->eventsToDispatch = [];
+    }
```
alexander-schranz, line 70: should also listen to `onClear` to clear the events if something went wrong or somebody did clear the unit of work.

Resources/config/services.xml
```diff
@@ -166,5 +166,13 @@
         </service>
 
         <service id="Sulu\Bundle\ContentBundle\Content\Application\ContentManager\ContentManagerInterface" alias="sulu_content.content_manager"/>
+
+        <!-- lazy is important because of the doctrine-event-listener, else this will result in a endless loop on cache warmup -->
```
alexander-schranz: sounds a little bit strange in which combination does this error occur? in combination with the `kernel.event_subscriber` tag or what does cause the problem?

Content/Infrastructure/Symfony/Workflow/FlushedEventSubscriber.php
wachterjohannes: we could hear on `workflow.completed` and check for the names (but they have to be declared then somehow)

Content/Infrastructure/Symfony/Workflow/FlushedEventSubscriber.php
wachterjohannes: 👍

Resources/config/services.xml
wachterjohannes: will recheck this - i had to add the lazy to the service in the project i have done the prototype! maybe here its not neccesary

Content/Infrastructure/Symfony/Workflow/FlushedEventSubscriber.php
wachterjohannes: do we know all the names of the worklows?

Resources/config/services.xml
wachterjohannes: not necceary and have removed it

PR #190, opened 2021-06-21
Add DimensionContentQueryEnhancer service

alexander-schranz:
The DimensionContentQueryEnhancer will make it possible the Repository of a ContentRichEntity in our current case the Article. The DimensionContentQueryEnhancer provide all required filters like categories, tags, template keys which are used by smart content. Also it provide the logic how the select join work as they are very complex to select only the dimensions which are required to be performant. The current implementation is very imperformant as when a list is loaded all dimension content are loaded as separate queries and are not correctly joined.

### DIFF
--- a/Content/Domain/Model/DimensionContentCollection.php
+++ b/Content/Domain/Model/DimensionContentCollection.php
@@ -59,7 +59,7 @@ class DimensionContentCollection implements \IteratorAggregate, DimensionContent
         $this->dimensionAttributes = $dimensionContentClass::getEffectiveDimensionAttributes($dimensionAttributes);
 
         $this->dimensionContents = new ArrayCollection(
-            // sort dimension content correctly by effective attributes for later merge
+            // dimension contents need to be sorted from most specific to least specific when they are merged
             SortUtils::multisort($dimensionContents, \array_keys($this->dimensionAttributes), 'asc')
         );
     }
--- a/Content/Domain/Model/DimensionContentCollection.php
+++ b/Content/Domain/Model/DimensionContentCollection.php
@@ -15,6 +15,7 @@ namespace Sulu\Bundle\ContentBundle\Content\Domain\Model;
 
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Criteria;
+use Sulu\Component\Util\SortUtils;
 
 /**
  * @implements \IteratorAggregate<DimensionContentInterface>
@@ -25,16 +26,6 @@ class DimensionContentCollection implements \IteratorAggregate, DimensionContent
      * @var ArrayCollection<int, DimensionContentInterface>
      */
     private $dimensionContents;
-
-    /**
-     * @var DimensionContentInterface|null
-     */
-    private $unlocalizedDimensionContent;
-
-    /**
-     * @var DimensionContentInterface|null
-     */
-    private $localizedDimensionContent;
 
     /**
      * @var mixed[]
@@ -63,23 +54,14 @@ class DimensionContentCollection implements \IteratorAggregate, DimensionContent
         array $dimensionAttributes,
         string $dimensionContentClass
     ) {
-        $this->dimensionContents = new ArrayCollection($dimensionContents);
         $this->dimensionContentClass = $dimensionContentClass;
         $this->defaultDimensionAttributes = $dimensionContentClass::getDefaultDimensionAttributes();
-
-        $this->unlocalizedDimensionContent = $this->dimensionContents->filter(
-            function(DimensionContentInterface $dimensionContent) {
-                return null === $dimensionContent->getLocale();
-            }
-        )->first() ?: null;
-
-        $this->localizedDimensionContent = $this->dimensionContents->filter(
-            function(DimensionContentInterface $dimensionContent) {
-                return null !== $dimensionContent->getLocale();
-            }
-        )->first() ?: null;
-
         $this->dimensionAttributes = $dimensionContentClass::getEffectiveDimensionAttributes($dimensionAttributes);
+
+        $this->dimensionContents = new ArrayCollection(
+            // sort dimension content correctly by effective attributes for later merge
+            SortUtils::multisort($dimensionContents, \array_keys($this->dimensionAttributes), 'asc')
+        );
     }
 
     public function getDimensionContentClass(): string
--- a/Content/Domain/Model/DimensionContentCollection.php
+++ b/Content/Domain/Model/DimensionContentCollection.php
@@ -79,7 +79,7 @@ class DimensionContentCollection implements \IteratorAggregate, DimensionContent
             }
         )->first() ?: null;
 
-        $this->dimensionAttributes = $dimensionAttributes;
+        $this->dimensionAttributes = $dimensionContentClass::getEffectiveDimensionAttributes($dimensionAttributes);
     }
 
     public function getDimensionContentClass(): string
--- a/Content/Domain/Model/DimensionContentInterface.php
+++ b/Content/Domain/Model/DimensionContentInterface.php
@@ -38,4 +38,11 @@ interface DimensionContentInterface
      * @return mixed[]
      */
     public static function getDefaultDimensionAttributes(): array;
+
+    /**
+     * @param mixed[] $dimensionAttributes
+     *
+     * @return mixed[]
+     */
+    public static function getEffectiveDimensionAttributes(array $dimensionAttributes): array;
 }
--- a/Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
+++ b/Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
@@ -36,10 +36,6 @@ class DimensionContentQueryEnhancer
     public const WITH_EXCERPT_TAGS = 'with-excerpt-tags';
     public const WITH_EXCERPT_CATEGORIES = 'with-excerpt-categories';
     public const WITH_EXCERPT_CATEGORIES_TRANSLATION = 'with-excerpt-categories-translation';
-    public const WITH_EXCERPT_IMAGE = 'with-excerpt-image';
-    public const WITH_EXCERPT_IMAGE_TRANSLATION = 'with-excerpt-image-translation';
-    public const WITH_EXCERPT_ICON = 'with-excerpt-icon';
-    public const WITH_EXCERPT_ICON_TRANSLATION = 'with-excerpt-icon-translation';
 
     /**
      * Groups are used in controllers and represents serialization / resolver group,
@@ -57,17 +53,11 @@ class DimensionContentQueryEnhancer
         self::GROUP_CONTENT_ADMIN => [
             self::WITH_EXCERPT_TAGS => true,
             self::WITH_EXCERPT_CATEGORIES => true,
-            self::WITH_EXCERPT_IMAGE => true,
-            self::WITH_EXCERPT_ICON => true,
         ],
         self::GROUP_CONTENT_WEBSITE => [
             self::WITH_EXCERPT_TAGS => true,
             self::WITH_EXCERPT_CATEGORIES => true,
             self::WITH_EXCERPT_CATEGORIES_TRANSLATION => true,
-            self::WITH_EXCERPT_IMAGE => true,
-            self::WITH_EXCERPT_IMAGE_TRANSLATION => true,
-            self::WITH_EXCERPT_ICON => true,
-            self::WITH_EXCERPT_ICON_TRANSLATION => true,
         ],
     ];
 
@@ -97,8 +87,10 @@ class DimensionContentQueryEnhancer
         $effectiveAttributes = $dimensionContentClassName::getEffectiveDimensionAttributes($filters);
 
         $queryBuilder->leftJoin(
-            $contentRichEntityAlias . '.dimensionContents',
-            'filterDimensionContent'
+            $dimensionContentClassName,
+            'filterDimensionContent',
+            Join::WITH,
+            'filterDimensionContent.' . $contentRichEntityAlias . ' = ' . $contentRichEntityAlias . ''
         );
 
         foreach ($effectiveAttributes as $key => $value) {
@@ -243,6 +235,7 @@ class DimensionContentQueryEnhancer
      */
     public function addSelects(
         QueryBuilder $queryBuilder,
+        string $contentRichEntityAlias,
         string $dimensionContentClassName,
         array $dimensionAttributes,
         array $selects = []
@@ -256,6 +249,11 @@ class DimensionContentQueryEnhancer
                 $selects = \array_merge($selects, self::SELECTS[$selectGroup]);
             }
         }
+
+        $queryBuilder->leftJoin(
+            $contentRichEntityAlias . '.dimensionContents',
+            'dimensionContent'
+        );
 
         $effectiveAttributes = $dimensionContentClassName::getEffectiveDimensionAttributes($dimensionAttributes);
         $this->addSortBy($queryBuilder, $effectiveAttributes);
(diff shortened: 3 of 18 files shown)


Content/Domain/Model/DimensionContentInterface.php
```diff
@@ -38,4 +38,11 @@ public function markAsMerged(): void;
      * @return mixed[]
      */
     public static function getDefaultDimensionAttributes(): array;
+
+    /**
+     * @param mixed[] $dimensionAttributes
+     *
+     * @return mixed[]
+     */
+    public static function getEffectiveDimensionAttributes(array $dimensionAttributes): array;
```
niklasnatter, line 47: not sure about this method as part of the interface.. i think the implementation will most likely be the same for all entities, right?

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
```diff
@@ -0,0 +1,223 @@
...
+     *     categoryIds?: int[],
+     *     categoryKeys?: string[],
+     *     categoryOperator?: 'AND'|'OR',
+     *     tagIds?: int[],
+     *     tagNames?: string[],
+     *     tagOperator?: 'AND'|'OR',
+     *     templateKeys?: string[],
+     * } $filters
+     */
+    public function enhanceDimensionContentFilter(
```
niklasnatter: the idea looks good but not sure about the naming yet 🙂 
could we merge the `enhanceDimensionContentFilter` and `enhanceDimensionContentSelect` method?

alexander-schranz: I don't think so. Filtering and selecting are different things (could be 2 different querybuilders) and the `Select` is not yet implemented.

Content/Domain/Model/DimensionContentInterface.php
alexander-schranz: I think they will stay the same, but I did not yet found a better place for them then beside the `getDefaultDimensionAttributes`. Another place would be the `DimensionContentCollection`.

niklasnatter: okay, i think i would at least change the name, not sure if `effective` is right here.. what do you think about `filterDimensionAttributes`/`getFilteredDimensionAttributes`/`getRelevantDimensionAttributes`?

niklasnatter: okay forget the comment above, i did not understand that this will merge the default values with the given attributes. in that case `effective` sounds good to me 👍

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
niklasnatter: okay i think i would just call this `addFilters`/`addDimensionContentFilters` then - what do you think? 🙂

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
```diff
@@ -0,0 +1,223 @@
...
+                sprintf('The operator "%s" is not supported for this filter.', $operator)
+            );
+        }
+    }
+
+    /**
+     * @param class-string<DimensionContentInterface> $dimensionContentClassName
+     * @param mixed[] $dimensionAttributes
+     */
+    public function enhanceDimensionContentSelect(
```
niklasnatter: do you use this function in the `SuluArticleBundle`?
if not, i think i would keep the logic in the `DimensionContentRepository` for now.
if yes, i think i would call it `addSelect`/`addEffectiveDimensionContentSelect` 🙂

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
alexander-schranz: sounds good for me.

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
alexander-schranz: It will be used in the article bundle, but think I need first implement it here in the Example Bundle to also write tests for it.

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
```diff
@@ -0,0 +1,229 @@
...
+     *     tagOperator?: 'AND'|'OR',
+     *     templateKeys?: string[],
+     * } $filters
+     */
+    public function addFilters(
+        QueryBuilder $queryBuilder,
+        string $dimensionContentClassName,
+        array $filters
+    ): void {
+        $rootAlias = $queryBuilder->getRootAliases()[0];
```
niklasnatter: maybe we should pass `contentRichEntityAlias` as argument?

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
```diff
@@ -0,0 +1,229 @@
...
+                sprintf('The operator "%s" is not supported for this filter.', $operator)
+            );
+        }
+    }
+
+    /**
+     * @param class-string<DimensionContentInterface> $dimensionContentClassName
+     * @param mixed[] $dimensionAttributes
+     */
+    public function addSelects(
```
niklasnatter, line 250: do you think there will be more selects beside of the `dimensionContent`?
think i would call this `addSelect` or `addDimensionContentSelect`

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
```diff
@@ -0,0 +1,229 @@
...
+        QueryBuilder $queryBuilder,
+        string $dimensionContentClassName,
+        array $dimensionAttributes
+    ): void {
+        $effectiveAttributes = $dimensionContentClassName::getEffectiveDimensionAttributes($dimensionAttributes);
+        $this->addSortBy($queryBuilder, $effectiveAttributes);
+        $queryBuilder->addCriteria($this->getAttributesCriteria('dimensionContent', $effectiveAttributes));
```
niklasnatter, line 267: maybe we should pass a `dimensionContentAlias` as parameter? 🤔

Tests/Application/ExampleTestBundle/Repository/ExampleRepository.php
```diff
@@ -0,0 +1,267 @@
...
+     */
+    public function findBy(array $filters = [], array $orderBy = []): iterable
+    {
+        $queryBuilder = $this->createQueryBuilder($filters, $orderBy);
+
+        /** @var iterable<Example> $examples */
+        $examples = $queryBuilder->getQuery()->getResult();
+
+        foreach ($examples as $example) {
+            yield $example;
```
niklasnatter, line 228: is there an advantage of yielding the results here instead of returning an array? 🤔

alexander-schranz: method should be refractored to us `toIterator` method of doctrine just need to test it. The advantage is a performance as if correctly used only 1 object instance is in memory why looping over it.

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
alexander-schranz: There will be a lot of more selects excerpt image, icon, categories, tags, ... depending on the controller or service loading it. So the `dimensionContent` will stay here hardcoded as it does not make sense to make that setable but the others not.

I also think that there would be some extension point needed. Example usecase could be the SuluDirectoryBundle extends the `Excerpt` Tab with directories and topics on all content able entities.

Tests/Application/ExampleTestBundle/Repository/ExampleRepository.php
alexander-schranz: Currently just used a custom Generator to make sure all works with a Generator.

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
alexander-schranz: There are many other joins so I would not introduce a parameter for the aliases here.

alexander-schranz:
@nnatter ready for review.

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
```diff
@@ -0,0 +1,333 @@
...
+                continue;
+            }
+
+            if (isset(self::SELECTS[$selectGroup])) {
+                $selects = \array_merge($selects, self::SELECTS[$selectGroup]);
+            }
+        }
+
+        $effectiveAttributes = $dimensionContentClassName::getEffectiveDimensionAttributes($dimensionAttributes);
+        $this->addSortBy($queryBuilder, $effectiveAttributes);
```
niklasnatter: i am not sure if i would expect this method to change my sorting. you want to sort this to return the most specific `DimensionContent` first, right?

niklasnatter: this also means that calling the `addSelects` method will lead to multiple rows per entity, right?

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
```diff
@@ -0,0 +1,333 @@
...
+        ContentMetadataInspectorInterface $contentMetadataInspector
+    ) {
+        $this->contentMetadataInspector = $contentMetadataInspector;
+    }
+
+    /**
+     * Withs represents additional selects which can be load to join and select specific sub entities.
+     * They are used by groups and fields.
+     */
+    public const WITH_EXCERPT_TAGS = 'with-excerpt-tags';
```
niklasnatter: what do you think about calling these `SELECT_*` and `SELECTGROUP_*`?

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
niklasnatter: i think i would add a `ALIAS_DIMENSION_CONTENT` constant then? 🙂

Content/Infrastructure/Doctrine/DimensionContentRepository.php
```diff
@@ -52,77 +57,24 @@ public function load(
 
         $queryBuilder = $this->entityManager->createQueryBuilder()
             ->from($dimensionContentClass, 'dimensionContent')
-            ->select('dimensionContent')
             ->innerJoin('dimensionContent.' . $mappingProperty, 'content')
```
niklasnatter, line 60: we could use the `DimensionContentQueryEnhancer::ALIAS_DIMENSION_CONTENT` here if we define one

Tests/Application/ExampleTestBundle/Repository/ExampleRepository.php
```diff
@@ -0,0 +1,381 @@
...
+
+            $this->dimensionContentQueryEnhancer->addSelects(
+                $queryBuilder,
+                ExampleDimensionContent::class,
+                $contentFilters,
+                $contentSelects
+            );
+        }
+
+        $locale = $dimensionAttributes['locale'] ?? null;
```
niklasnatter: i think `$dimensionAttributes` is not defined anywhere? wondering why phpstan is okay with this? 🙈

Tests/Application/ExampleTestBundle/Repository/ExampleRepository.php
```diff
@@ -0,0 +1,381 @@
...
+        if (null !== $page) {
+            Assert::notNull($limit);
+            Assert::integer($page);
+            $offset = (int) ($limit * ($page - 1));
+            $queryBuilder->setFirstResult($offset);
+        }
+
+        $contentFilters = [];
+        foreach ([
+            'locale',
```
niklasnatter: this would be needed in every repository and makes it harder to extend things 😕

niklasnatter: maybe we can just pass all `$filters` to the `dimensionContentQueryEnhancer`?

niklasnatter: or we could add a prefix for filters that should be passed to the `dimensionContentQueryEnhancer`? eg `dimensionContent.categoryIds`

Tests/Application/ExampleTestBundle/Repository/ExampleRepository.php
```diff
@@ -0,0 +1,381 @@
...
+        if ($selects['with-example-content'] ?? null) {
+            /** @var array<string, bool> $contentSelects */
+            $contentSelects = $selects['with-example-content'] ?? [];
+
+            $queryBuilder->leftJoin(
+                'example.dimensionContents',
+                'dimensionContent'
+            );
+
+            $this->dimensionContentQueryEnhancer->addSelects(
```
niklasnatter, line 325: this will lead to multiple rows for a single example like described in https://github.com/sulu/SuluContentBundle/pull/190#discussion_r712297997, right? 🤔

alexander-schranz: As discussed addSelect will just add the joined to relation and not create a new instance of the example object.

Tests/Application/ExampleTestBundle/Repository/ExampleRepository.php
alexander-schranz: this is false, funny that the test doesn't cover this

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
alexander-schranz: Will remove this sorting. in the future maybe it make sense that we can sort specially for smart content by published date but that should be handled in a additional addSortBy method.

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
alexander-schranz: I would not yet add any constant here for aliases.

Tests/Application/ExampleTestBundle/Repository/ExampleRepository.php
alexander-schranz: I'm fine currently to forward the filters directly to the enhancer

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
alexander-schranz: As discussed is fine for me to to use a flat list with `SELECT_` and `SELECTGROUP_` constants.

Content/Infrastructure/Doctrine/DimensionContentRepository.php
alexander-schranz: as discussed we will currently avoid the alias constants as it hard to define which will and which will not be constants ad I think it will not improve the DX

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
alexander-schranz: should we go with `SELECT_GROUP_` as the IDE will complain else always about the word. Or with `GROUP_SELECT_..` to have all SELECT and all GROUP together in autocomplete?

alexander-schranz: I did go with `GROUP_SELECT_` what do you think?

Tests/Application/ExampleTestBundle/Repository/ExampleRepository.php
alexander-schranz: did update it.

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
niklasnatter: `GROUP_SELECT_ ` sounds good to me 🙂

Content/Domain/Model/DimensionContentCollection.php
```diff
@@ -26,16 +27,6 @@ class DimensionContentCollection implements \IteratorAggregate, DimensionContent
      */
     private $dimensionContents;
 
-    /**
-     * @var DimensionContentInterface|null
-     */
-    private $unlocalizedDimensionContent;
```
alexander-schranz, line 32: this are private variables which were just set but not used.

alexander-schranz:
@nnatter Ready for rereview ;)

Content/Domain/Model/DimensionContentCollection.php
```diff
@@ -63,23 +54,14 @@ public function __construct(
...
-
-        $this->localizedDimensionContent = $this->dimensionContents->filter(
-            function(DimensionContentInterface $dimensionContent) {
-                return null !== $dimensionContent->getLocale();
-            }
-        )->first() ?: null;
-
-        $this->dimensionAttributes = $dimensionAttributes;
+        $this->dimensionContents = new ArrayCollection(
+            // sort dimension content correctly by effective attributes for later merge
```
niklasnatter:
```suggestion
            // dimension contents need to be sorted from most specific to least specific when they are merged
```

Tests/Application/ExampleTestBundle/Repository/ExampleRepository.php
```diff
@@ -0,0 +1,347 @@
...
+use Webmozart\Assert\Assert;
+
+class ExampleRepository
+{
+    /**
+     * Groups are used in controllers and represents serialization / resolver group,
+     * this allows that no controller need to be overwritten when something additional should be
+     * loaded at that endpoint.
+     */
+    public const GROUP_EXAMPLE_ADMIN = 'example_admin';
```
niklasnatter: would be nice to use `SELECT_GROUP_` here too

Tests/Application/ExampleTestBundle/Repository/ExampleRepository.php
```diff
@@ -0,0 +1,347 @@
...
+    public const GROUP_EXAMPLE_WEBSITE = 'example_website';
+
+    /**
+     * Withs represents additional selects which can be load to join and select specific sub entities.
+     * They are used by groups.
+     */
+    public const WITH_EXAMPLE_CONTENT = 'with-example-content';
```
niklasnatter: would be nice to use `SELECT_` here too

Tests/Application/ExampleTestBundle/Repository/ExampleRepository.php
alexander-schranz: you mean `GROUP_SELECT_` :)

Tests/Application/ExampleTestBundle/Repository/ExampleRepository.php
alexander-schranz: updated

PR #191, opened 2021-06-22
Fix phpstan errors

alexander-schranz:


.github/workflows/test-application.yaml
```diff
@@ -22,26 +22,20 @@ jobs:
             matrix:
                 include:
                     - php-version: '7.2'
-                      lint: false
                       coverage: false
                       dependency-versions: 'lowest'
-                      tools: 'composer:v1'
```
alexander-schranz, line 28: Aslong as we don't have a composer plugin I see no need to use composer:v1 in any of our CI's it just cost us performance and take longer to install.

PR #194, opened 2021-09-14
Upgrade php-cs-fixer

alexander-schranz:


.php-cs-fixer.dist.php
```diff
@@ -9,33 +9,32 @@
 with this source code in the file LICENSE.
 EOF;
 
-return PhpCsFixer\Config::create()
-    ->setRiskyAllowed(true)
+$finder = PhpCsFixer\Finder::create()
```
niklasnatter: i think this is not used? 🤔

alexander-schranz: good catch 👍

niklasnatter (approved):

PR #198, opened 2021-09-28
Add subscriber to delete route when a content-rich entity got removed

wachterjohannes:
fixes #198


Content/Infrastructure/Doctrine/RouteRemover.php
```diff
@@ -0,0 +1,52 @@
...
+namespace Sulu\Bundle\ContentBundle\Content\Infrastructure\Doctrine;
+
+use Doctrine\Common\EventSubscriber;
+use Doctrine\Common\Util\ClassUtils;
+use Doctrine\ORM\Event\LifecycleEventArgs;
+use Doctrine\ORM\Events;
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\ContentRichEntityInterface;
+use Sulu\Bundle\RouteBundle\Entity\RouteRepositoryInterface;
+
+class RouteRemover implements EventSubscriber
```
alexander-schranz, line 27: This should not be a public service so it should be internal and final:

```suggestion
/**
 * @internal
 * @final
 */
class RouteRemover implements EventSubscriber
```

Tests/Unit/Content/Infrastructure/Doctrine/RouteRemoverTest.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    public function testPreRemove(): void
+    {
+        $object = $this->prophesize(ContentRichEntityInterface::class);
+        $object->getId()->willReturn('123-123-123');
+
+        $entityManager = $this->prophesize(EntityManagerInterface::class);
+
+        $event = $this->prophesize(LifecycleEventArgs::class);
```
alexander-schranz: would not mock the `LifecycleEventArgs` we should go with a real event instance:

```
$event = new LifecycleEventArgs($object, $entityManager->reveal());
```

alexander-schranz (changes requested):

Content/Infrastructure/Doctrine/RouteRemover.php
wachterjohannes: done

Tests/Unit/Content/Infrastructure/Doctrine/RouteRemoverTest.php
wachterjohannes: done

Content/Infrastructure/Doctrine/RouteRemover.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    public function preRemove(LifecycleEventArgs $event): void
+    {
+        $object = $event->getObject();
+        if (!$object instanceof ContentRichEntityInterface) {
+            return; // @codeCoverageIgnore
+        }
+
+        $entityClass = ClassUtils::getRealClass(\get_class($object));
```
niklasnatter: why do we need `ClassUtils::getRealClass` here? 🤔 
i think we should determine the class based on the resource key like we do in the `RoutableDataMapper`:
https://github.com/sulu/SuluContentBundle/blob/bd6674a20de5737a35a3fdfab373a1e495cf75ac/Content/Application/ContentDataMapper/DataMapper/RoutableDataMapper.php#L140-L147

we should be able to determine the class that implements the `RoutableInterface` using the `ContentMetadataInspector:: getDimensionContentClass`

Content/Infrastructure/Doctrine/RouteRemover.php
```diff
@@ -0,0 +1,56 @@
...
+
+    public function getSubscribedEvents()
+    {
+        return [Events::preRemove];
+    }
+
+    public function preRemove(LifecycleEventArgs $event): void
+    {
+        $object = $event->getObject();
+        if (!$object instanceof ContentRichEntityInterface) {
```
niklasnatter, line 65: wouldnt it be more intuitive to do this for `RoutableInterface` entities? i guess you did not do that because they are deleted because of database level cascading, right?

wachterjohannes: exactly - they are moved by cascading on database level

Content/Infrastructure/Doctrine/RouteRemover.php
wachterjohannes: we would have to load unnecessary data from the database (dimensions) - The `ClassUtils::getRealClass` removes the proxy classes from the return - so we get the correct class

alexander-schranz: The `ContentMetadataInspector` does not load anything.

alexander-schranz: @wachterjohannes still would use the `ContentMetadataInspector` here and then use ClassUtils::getRealClass inside it. But we should use here the routeMappings configuration.

wachterjohannes: @alexander-schranz is there a good reason todo it? it is really don't necessary and i would keep it as slim as possible

alexander-schranz: It should be consistent and the only source of true for the route configuraiton is the routingMapping variable. We had in the past problems when somebody did overwrite entities that it did use get_class and it did then not longer work.

wachterjohannes:
@alexander-schranz @nnatter fixed

Tests/Application/ExampleTestBundle/Entity/Example.php
```diff
@@ -37,10 +37,18 @@ public function getId()
         return $this->id;
     }
 
-    public function createDimensionContent(): DimensionContentInterface
+    /**
+     * @param mixed $id
+     */
+    public function setId($id): self
```
niklasnatter: i am afraid that people will blindly copy (and maybe even use) this if we add it to the example.
we only need this for the test - right? i think i would set it via reflection in the test then

alexander-schranz (approved):

Tests/Application/ExampleTestBundle/Entity/Example.php
wachterjohannes: @nnatter done

niklasnatter (approved):

PR #200, opened 2021-09-30
Fix kernel and preview configuration fr example preview

wachterjohannes:


Tests/Application/PreviewKernelFactory.php
```diff
@@ -0,0 +1,28 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\ContentBundle\Tests\Application;
+
+use Sulu\Bundle\PreviewBundle\Preview\Renderer\KernelFactoryInterface;
+use Sulu\Component\HttpKernel\SuluKernel;
+
+class PreviewKernelFactory implements KernelFactoryInterface
```
alexander-schranz, line 19: Is this still required when setting the class alias?

wachterjohannes: yes - else the wrong project dir is used - see the overwritten class

alexander-schranz: But is then the Alias still required?

Tests/Application/PreviewKernel.php
```diff
@@ -0,0 +1,24 @@
...
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\ContentBundle\Tests\Application;
+
+use Sulu\Bundle\PreviewBundle\Preview\Renderer\PreviewKernel as SuluPreviewKernel;
+
+class PreviewKernel extends SuluPreviewKernel
+{
+    public function getProjectDir(): string
```
alexander-schranz, line 20: I think we could set this directly on the `Tests/Application/Kernel.php` or why the preview should have another ProjectDir then the real kernel?

wachterjohannes: we cannot set it because the kernel itself has no property projectDir only the PreviewKernel has one and so it detects the project dir by searching for the composer.json

Tests/Application/PreviewKernelFactory.php
wachterjohannes: yes because i use the PreviewKernel to extend and the PreviewKernel extends `App\Kernel`

Tests/Application/PreviewKernel.php
alexander-schranz: okay we should in future make sure the projectdir is not longer used as its deprecated but at current state fine for me as it only has effects on the test setup.

alexander-schranz (approved):

PR #201, opened 2021-09-30
Add usage of tag 'sulu.rlp' to detect route property

wachterjohannes:


alexander-schranz:
The `sulu.rlp` tag is [deprecated](https://github.com/sulu/sulu/blob/2.x/UPGRADE.md#sulurlp-tag-deprecated) and should should not longer be used for this kind of things. Route attribute should only be loaded by the route types `route`, `resourceLocator`, ... not by a tag.

PR #202, opened 2021-10-02
Add versioning to dimension content

alexander-schranz:
Build on top of: #190

To be discussed:

 - [ ] Optimize getPublishLocales on write?
 - [ ] Optimize getLatestVersion on write?
 - [ ] Versioning an own Interface (make getEffectiveDimensionAttributes and getDefaultDimensionAttributes more complicated is it worse?)
 - [ ] Is Restore version a workflow transition?
 - [ ] Should Version be an own Entity / Table extending from other entity? (complexity on overwrite entity (could be validated in compilerpass))
 
 TODO:
 
 - [ ] UI Versioning list on settings
 - [ ] Restore Version

### DIFF
--- a/Content/Application/ContentWorkflow/ContentWorkflow.php
+++ b/Content/Application/ContentWorkflow/ContentWorkflow.php
@@ -135,9 +135,11 @@ class ContentWorkflow implements ContentWorkflowInterface
         // | New |--------->| Unpublished |                      | Review |---------->| Published  |                | draft |                             | Review draft  |
         // |     |          |             |<---------------------|        |           |            |--------------->|       |<----------------------------|               |
         // +-----+          +-------------+       reject         +--------+           +------------+  create draft  +-------+        reject draft         +---------------+
-        //                                                                              A   |    A                                                                |
-        //                                                                              +---+    |                          publish                               |
-        //                                                                             publish   +----------------------------------------------------------------+
+        //                   A   |                                                     A   |  A  |                      A                                          |
+        //                   +---+                                                     +---+  |  |       restore        |                                          |
+        //                  restore                                                   publish |  +----------------------+                                          |
+        //                                                                                    |                              publish                               |
+        //                                                                                    +--------------------------------------------------------------------+
 
         // Configures places
         $definition = $definitionBuilder
--- a/Content/Application/ContentWorkflow/Subscriber/PublishTransitionSubscriber.php
+++ b/Content/Application/ContentWorkflow/Subscriber/PublishTransitionSubscriber.php
@@ -19,6 +19,7 @@ use Sulu\Bundle\ContentBundle\Content\Domain\Model\ContentRichEntityInterface;
 use Sulu\Bundle\ContentBundle\Content\Domain\Model\DimensionContentCollectionInterface;
 use Sulu\Bundle\ContentBundle\Content\Domain\Model\DimensionContentInterface;
 use Sulu\Bundle\ContentBundle\Content\Domain\Model\WorkflowInterface;
+use Sulu\Bundle\ContentBundle\Content\Domain\Repository\DimensionContentRepositoryInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\Workflow\Event\TransitionEvent;
 
@@ -29,9 +30,17 @@ class PublishTransitionSubscriber implements EventSubscriberInterface
      */
     private $contentCopier;
 
-    public function __construct(ContentCopierInterface $contentCopier)
-    {
+    /**
+     * @var DimensionContentRepositoryInterface
+     */
+    private $dimensionContentRepository;
+
+    public function __construct(
+        ContentCopierInterface $contentCopier,
+        DimensionContentRepositoryInterface $dimensionContentRepository
+    ) {
         $this->contentCopier = $contentCopier;
+        $this->dimensionContentRepository = $dimensionContentRepository;
     }
 
     public function onPublish(TransitionEvent $transitionEvent): void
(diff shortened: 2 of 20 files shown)


Content/Application/ContentWorkflow/ContentWorkflow.php
```diff
@@ -135,9 +135,11 @@ private function getWorkflow(): SymfonyWorkflowInterface
...
-        //                                                                              +---+    |                          publish                               |
-        //                                                                             publish   +----------------------------------------------------------------+
+        //                   A   |                                                     A   |  A  |                      A                                          |
+        //                   +---+                                                     +---+  |  |       restore        |                                          |
+        //                  restore                                                   publish |  +----------------------+                                          |
+        //                                                                                    |                              publish                               |
+        //                                                                                    +--------------------------------------------------------------------+
```
alexander-schranz, line 142: See [whole Graphic](https://github.com/sulu/SuluContentBundle/blob/651aa2ae8ac7d0f450b782ad24a0a75e3c623a49/Content/Application/ContentWorkflow/ContentWorkflow.php#L124-L142)

/cc @wachterjohannes not sure if even possible have a transition with the same name with different source and target state?

```
published     ----restore---> draft
unpublished   ----restore---> unpublished
```

wachterjohannes: > Is Restore version a workflow transition?

Answer in my opinion - yes - it should be if possible

> if even possible have a transition with the same name with different source and target state?

Yes it is for multiple sources - but as i know only to a single destination. So your example i dont think its possible (maybe with a guard or a listener which sets the destination dynamically by using the further state?)

Content/Application/ContentWorkflow/Subscriber/PublishTransitionSubscriber.php
```diff
@@ -68,6 +77,21 @@ public function onPublish(TransitionEvent $transitionEvent): void
 
         $dimensionAttributes['stage'] = DimensionContentInterface::STAGE_LIVE;
 
+        // create new version
+        // TODO optimize latest version and publish locales on write process to avoid loading them here?
```
wachterjohannes, line 81: 👍 we should optimize this information

Content/Application/ContentWorkflow/Subscriber/PublishTransitionSubscriber.php
```diff
@@ -68,6 +77,21 @@ public function onPublish(TransitionEvent $transitionEvent): void
...
+        $version = 1 + $this->dimensionContentRepository->getLatestVersion($contentRichEntity);
+        $publishLocales = $this->dimensionContentRepository->getLocales($contentRichEntity, $dimensionAttributes);
+
+        foreach ($publishLocales as $publishLocale) {
+            $this->contentCopier->copy(
+                $contentRichEntity,
+                \array_merge($dimensionAttributes, ['locale' => $publishLocale]),
+                $contentRichEntity,
+                \array_merge($dimensionAttributes, ['locale' => $publishLocale, 'version' => $version])
```
wachterjohannes, line 90: 👍 nice solution with the version as attribute

Content/Domain/Model/DimensionContentCollection.php
```diff
@@ -63,23 +54,14 @@ public function __construct(
...
         $this->defaultDimensionAttributes = $dimensionContentClass::getDefaultDimensionAttributes();
+        $this->dimensionAttributes = $dimensionContentClass::getEffectiveDimensionAttributes($dimensionAttributes);
 
-        $this->unlocalizedDimensionContent = $this->dimensionContents->filter(
-            function(DimensionContentInterface $dimensionContent) {
-                return null === $dimensionContent->getLocale();
-            }
-        )->first() ?: null;
-
-        $this->localizedDimensionContent = $this->dimensionContents->filter(
```
wachterjohannes: are they not used anymore?

Content/Domain/Model/DimensionContentTrait.php
```diff
@@ -65,6 +80,22 @@ public static function getDefaultDimensionAttributes(): array
         return [
             'locale' => null,
             'stage' => DimensionContentInterface::STAGE_DRAFT,
+            'version' => DimensionContentInterface::DEFAULT_VERSION,
         ];
     }
+
+    public static function getEffectiveDimensionAttributes(array $dimensionAttributes): array
```
wachterjohannes: hmm - why effective? why do you need that?

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
```diff
@@ -0,0 +1,322 @@
...
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\TemplateInterface;
+use Webmozart\Assert\Assert;
+
+/**
+ * TODO add loadGhost functionality
+ *      add loadShadow functionality.
+ *
+ * @final
+ */
+class DimensionContentQueryEnhancer
```
wachterjohannes: isnt that part of a different PR?

Tests/Application/ExampleTestBundle/Repository/ExampleRepository.php
```diff
@@ -0,0 +1,347 @@
...
+    public const WITH_EXAMPLE_CONTENT = 'with-example-content';
+
+    public const WITH_EXAMPLE_TRANSLATION = 'with-example-translation';
+
+    /**
+     * TODO it should be possible to extend fields and groups inside the SELECTS.
+     */
+    private const SELECTS = [
+        // GROUPS
+        self::GROUP_EXAMPLE_ADMIN => [
```
wachterjohannes: should this realy be needed? that is always the same isnt it? at least as long as there is no real extension

wachterjohannes (changes requested):
am i right when i say the restore itself is not build until now? only the save to a new version during publishing?

wachterjohannes:
>  Optimize getPublishLocales on write?

Yes

> Optimize getLatestVersion on write?

Yes

> Versioning an own Interface (make getEffectiveDimensionAttributes and getDefaultDimensionAttributes more complicated is it worse?)

Would say no - as long as the implementation of the custom entity wont get to complex (doesnt look like as the moment) at the end this should be a basic and integrated feature for all entities.

> Is Restore version a workflow transition?

Already answered in the review above.

> Should Version be an own Entity / Table extending from other entity? (complexity on overwrite entity (could be validated in compilerpass))

Would say no - But we should discuss that with @chirimoya because of the problem that the table can get really big (bigger than it is already). But that problem we also have with the current phpcr implementation (at least with doctrine dbal storage)

Content/Domain/Model/DimensionContentCollection.php
alexander-schranz: This was already part of  #190

Content/Domain/Model/DimensionContentTrait.php
alexander-schranz: This was already part of  #190

Content/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
alexander-schranz: yes but this PR was build on top of that ;)

Tests/Application/ExampleTestBundle/Repository/ExampleRepository.php
alexander-schranz: This was already part of  #190

PR #205, opened 2021-10-06
Simplify data mapper with unlocalized and localized dimension content

alexander-schranz:
For making the mapper easier to handle and avoid check for localized and unlocalized dimension I for changing the interface to the following way.

TODO:

 - [x] UPGRADE.md DataMapperInterface

### DIFF
--- a/Content/Application/ContentDataMapper/ContentDataMapper.php
+++ b/Content/Application/ContentDataMapper/ContentDataMapper.php
@@ -32,11 +32,23 @@ class ContentDataMapper implements ContentDataMapperInterface
     }
 
     public function map(
-        array $data,
-        DimensionContentCollectionInterface $dimensionContentCollection
+        DimensionContentCollectionInterface $dimensionContentCollection,
+        array $dimensionAttributes,
+        array $data
     ): void {
+        $localizedDimensionAttributes = $dimensionAttributes;
+        $unlocalizedDimensionAttributes = $dimensionAttributes;
+        $unlocalizedDimensionAttributes['locale'] = null;
+        $unlocalizedDimensionContent = $dimensionContentCollection->getDimensionContent($unlocalizedDimensionAttributes);
+        $localizedDimensionContent = $dimensionContentCollection->getDimensionContent($localizedDimensionAttributes);
+
+        if (!$unlocalizedDimensionContent ||!$localizedDimensionContent) {
+            // TODO see https://github.com/sulu/SuluContentBundle/pull/204
+            throw new \RuntimeException('Create unlocalized and localized dimension content.');
+        }
+
         foreach ($this->dataMappers as $mapper) {
-            $mapper->map($data, $dimensionContentCollection);
+            $mapper->map($data, $unlocalizedDimensionContent, $localizedDimensionContent);
         }
     }
 }
--- a/Content/Application/ContentDataMapper/ContentDataMapper.php
+++ b/Content/Application/ContentDataMapper/ContentDataMapper.php
@@ -48,7 +48,7 @@ class ContentDataMapper implements ContentDataMapperInterface
         }
 
         foreach ($this->dataMappers as $mapper) {
-            $mapper->map($data, $unlocalizedDimensionContent, $localizedDimensionContent);
+            $mapper->map($unlocalizedDimensionContent, $localizedDimensionContent, $data);
         }
     }
 }
--- a/Content/Application/ContentDataMapper/ContentDataMapper.php
+++ b/Content/Application/ContentDataMapper/ContentDataMapper.php
@@ -42,7 +42,7 @@ class ContentDataMapper implements ContentDataMapperInterface
         $unlocalizedDimensionContent = $dimensionContentCollection->getDimensionContent($unlocalizedDimensionAttributes);
         $localizedDimensionContent = $dimensionContentCollection->getDimensionContent($localizedDimensionAttributes);
 
-        if (!$unlocalizedDimensionContent ||!$localizedDimensionContent) {
+        if (!$unlocalizedDimensionContent || !$localizedDimensionContent) {
             // TODO see https://github.com/sulu/SuluContentBundle/pull/204
             throw new \RuntimeException('Create unlocalized and localized dimension content.');
         }
--- a/Content/Application/ContentDataMapper/DataMapper/RoutableDataMapper.php
+++ b/Content/Application/ContentDataMapper/DataMapper/RoutableDataMapper.php
@@ -17,7 +17,6 @@ use Sulu\Bundle\ContentBundle\Content\Domain\Model\DimensionContentInterface;
 use Sulu\Bundle\ContentBundle\Content\Domain\Model\RoutableInterface;
 use Sulu\Bundle\ContentBundle\Content\Domain\Model\TemplateInterface;
 use Sulu\Bundle\RouteBundle\Generator\RouteGeneratorInterface;
-use Sulu\Bundle\RouteBundle\Manager\ConflictResolverInterface;
 use Sulu\Bundle\RouteBundle\Manager\RouteManagerInterface;
 use Sulu\Component\Content\Metadata\Factory\StructureMetadataFactoryInterface;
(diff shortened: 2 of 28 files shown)


Content/Application/ContentDataMapper/ContentDataMapper.php
```diff
@@ -32,11 +32,23 @@ public function __construct(iterable $dataMappers)
     }
 
     public function map(
-        array $data,
-        DimensionContentCollectionInterface $dimensionContentCollection
+        DimensionContentCollectionInterface $dimensionContentCollection,
+        array $dimensionAttributes,
```
niklasnatter, line 36: the `DimensionContentCollection` has a `getDimensionAttributes` method at the moment. will the attributes pass here be different from the ones returned by the method?

Content/Application/ContentDataMapper/DataMapper/RoutableDataMapper.php
```diff
@@ -147,44 +135,43 @@ public function map(
         }
 
         if (null === $entityClass || null === $routeSchema) {
-            // TODO FIXME add test case for this
-            return; // @codeCoverageIgnore
+            return;
```
niklasnatter: i think this is most likely an error (eg forget to add configuration). i think i would throw an exception here too?

Content/Application/ContentDataMapper/ContentDataMapper.php
alexander-schranz: This will be changed. The `DimensionContentCollection` future `Content` could have loaded multiple locales, which can then be mapped and so the mapper need to be given which data it should map to content to.

Tests/Application/ExampleTestBundle/Controller/ExampleController.php
```diff
@@ -136,6 +136,10 @@ public function postAction(Request $request): Response
     {
         $example = new Example();
 
+        $this->entityManager->persist($example);
+        // FIXME This flush is needed for RoutableMapper and should in future note longer be required
```
niklasnatter: why was this not needed before? 🤔

Tests/Application/ExampleTestBundle/Entity/ExampleDimensionContent.php
```diff
@@ -85,14 +85,18 @@ public function setTitle(?string $title): void
     public function getTemplateData(): array
     {
         $data = $this->parentGetTemplateData();
-        $data['title'] = $this->getTitle();
+        $title = $this->getTitle();
+
+        if ($title) {
+            $data['title'] = $title;
```
niklasnatter: is there a reason for doing this? 🤔 
i think the api for pages returns null values too

Tests/Application/ExampleTestBundle/Controller/ExampleController.php
alexander-schranz: We did skip when no ResourceId was given, should we readd this?

Tests/Functional/Integration/ExampleControllerTest.php
```diff
@@ -140,7 +140,7 @@ public function testPost(): int
         $this->assertResponseSnapshot('example_post.json', $response, 201);
 
         $routeRepository = $this->getContainer()->get('sulu.repository.route');
-        $this->assertCount(0, $routeRepository->findAll());
+        $this->assertCount(1, $routeRepository->findAll());
```
niklasnatter: 😮

Tests/Application/ExampleTestBundle/Entity/ExampleDimensionContent.php
alexander-schranz: Would make most tests confusing when its always returned here.

Tests/Functional/Integration/ExampleControllerTest.php
alexander-schranz: That is why the flush is now required for the ExampleController. That a route is generated. I added a todo that the RouteBundle should have  a way to generate a route for entities which do not have yet an id and the route should in that case be inserted on post flush then.

UPGRADE.md
```diff
@@ -1,5 +1,53 @@
 # Upgrade
 
+## 0.7.0
+
+### ContentMapperInterface changed
+
+The `ContentMapperInterface` has changed.
```
niklasnatter:
```suggestion
The `ContentMapperInterface` was changed as a preparation for refactoring the `DimensionContentCollection`:
```

UPGRADE.md
```diff
@@ -1,5 +1,53 @@
...
+    public function map(
+        DimensionContentCollectionInterface $dimensionContentCollection,
+        array $dimensionAttributes,
+        array $data
+    ): void;
+```
+
+### DataMapperInterface changed
+
+The `DataMapperInterface` has changed to make the mapper easier to set localized and unlocalized data.
```
niklasnatter:
```suggestion
The `DataMapperInterface` was changed to make it easier to set localized and unlocalized data:
```

Content/Application/ContentDataMapper/ContentDataMapper.php
niklasnatter: alright, sounds good 🙂

Tests/Application/ExampleTestBundle/Controller/ExampleController.php
niklasnatter: okay i see 🙂  i am not sure..
routes for pages are created only when they are published - maybe we should do that here too?

Tests/Functional/Integration/ExampleControllerTest.php
niklasnatter: yes i figured that thats the reason 🙂

Tests/Application/ExampleTestBundle/Entity/ExampleDimensionContent.php
niklasnatter: what test cases? 🤔

Tests/Application/ExampleTestBundle/Controller/ExampleController.php
niklasnatter: but i dont have a strong opinion on this, we dont need to change it now

Tests/Application/ExampleTestBundle/Entity/ExampleDimensionContent.php
alexander-schranz: Testcases using like Routablemapper using `getTemplateData()` but don't have and `title` in there mapping.

alexander-schranz: I changed the handling here by writing the 'title'  into the `TemplateData` in the `setTemplateData` and the only way to set the title is now `setTemplateData`.

alexander-schranz: So `getTemplateData` will only return a title when it really did set a title and not magically return a title when never was set.

Tests/Application/ExampleTestBundle/Controller/ExampleController.php
alexander-schranz: You are definitly right we currently have bug in the `RoutableDataMapper` after the entity is once published. The draft route edit is directly editing without ever being published.

Content/Application/ContentDataMapper/DataMapper/RoutableDataMapper.php
```diff
@@ -80,6 +80,10 @@ public function map(
             throw new \RuntimeException('LocalizedDimensionContent needs to extend the TemplateInterface.');
         }
 
+        if (DimensionContentInterface::STAGE_LIVE !== $localizedDimensionContent->getStage()) {
+            // return;
```
niklasnatter, line 84: this condition does not do anything anymore - is okay for me to merge it like this as we decided to refactor this part of the code in a second pull request 🙂

alexander-schranz: Fixed in #206

PR #209, opened 2021-10-13
Fix object provider and route defaults provider without id

alexander-schranz:
If an entity has not an id instead of uuid as identifier the query in content object provider should not fail. In future the entity for `smart_content`, `preview`, `route default provider` and `workflow` should be provided over a `ContentObjectProvider` and not by a query build inside the content bundle.


wachterjohannes (changes requested):
@alexander-schranz makes sense but seems that it does not work 🙈

wachterjohannes (approved):

PR #210, opened 2021-10-13
Fix set authored and author to null

alexander-schranz:
Currently errors but also the author and authored can be null and should be able to be set to null. There are projects where author is optional so this fields should be nullable and so removeable and setable by the content manager. Fixes https://github.com/sulu/sulu/pull/4026


wachterjohannes (changes requested):
@alexander-schranz Code looks good - but all pipelines failing :)

wachterjohannes (approved):

PR #211, opened 2021-10-13
Add Webspace select to Settings Tab

alexander-schranz:
Most entities like `Article` and any routable Custom Entity will have a webspace selection in its settings tab. I did go here with the name `WebspaceInterface`.

The entity like `Page` will not implement it as the webspace will there not be on the PageDimensionContent Entity and so not implement for its webspace a content interface. Other entity like snippet which will have content will not have a Webspace field or relation.

## TODO

 - [x] Move additional webspaces to seperate PR: (https://github.com/sulu/SuluContentBundle/pull/213)
 - [x] Add issue for SEO Canonical Handling (https://github.com/sulu/SuluContentBundle/issues/214)

### DIFF
--- a/Resources/config/data-mapper.xml
+++ b/Resources/config/data-mapper.xml
@@ -23,6 +23,12 @@
         </service>
 
         <service id="sulu_content.workflow_data_mapper" class="Sulu\Bundle\ContentBundle\Content\Application\ContentDataMapper\DataMapper\WorkflowDataMapper">
+            <tag name="sulu_content.data_mapper" priority="24"/>
+        </service>
+
+        <service id="sulu_content.webspace_data_mapper" class="Sulu\Bundle\ContentBundle\Content\Application\ContentDataMapper\DataMapper\WebspaceDataMapper">
+            <argument type="service" id="sulu_core.webspace.webspace_manager"/>
+
             <tag name="sulu_content.data_mapper" priority="16"/>
         </service>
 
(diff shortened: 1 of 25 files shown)


Resources/config/data-mapper.xml
```diff
@@ -23,6 +23,12 @@
         </service>
 
         <service id="sulu_content.workflow_data_mapper" class="Sulu\Bundle\ContentBundle\Content\Application\ContentDataMapper\DataMapper\WorkflowDataMapper">
+            <tag name="sulu_content.data_mapper" priority="24"/>
```
wachterjohannes, line 26: why 24? seems a strange number 🙈

Resources/config/merger.xml
```diff
@@ -17,6 +17,10 @@
         </service>
 
         <service id="sulu_content.seo_merger" class="Sulu\Bundle\ContentBundle\Content\Application\ContentMerger\Merger\SeoMerger">
+            <tag name="sulu_content.merger" priority="24"/>
```
wachterjohannes, line 20: same here

wachterjohannes (changes requested):

Resources/config/merger.xml
alexander-schranz: is between the other mergers

Resources/config/data-mapper.xml
alexander-schranz: @wachterjohannes Which steps between the mergers would you use?

wachterjohannes: its ok 😂 thought it is a strange number - maybe add a comment that it has to be between

wachterjohannes (approved):

PR #213, opened 2021-10-14
Add Additional Webspaces Field to Settings Tab

alexander-schranz:
Most entities like `Article` and any routable Custom Entity will have a webspace selection in its settings tab. I did go here with the name `WebspaceInterface`.

The entity like `Page` will not implement it as the webspace will there not be on the PageDimensionContent Entity and so not implement for its webspace a content interface. Other entity like snippet which will have content will not have a Webspace field or relation.

## TODO

 - [ ] Discuss how `additionalWebspaces` are saved
    - [ ] Version A: JSON Field (performant smart content query)
    - [ ] Version B: Additional Table with Webspace (primaryKey is `webspaceKey`)
    - [ ] Version C: Additional General Table (DimensionContentWebspace entity)
    - [ ] Version D: Additional Table without ORM (not sure if possible)
    
See prototype repository: https://github.com/alexander-schranz/doctrine-json-array-prototype/pull/1
    
Keep in mind for `navigationContexts` we need something similar, but as they are Page specific and will not be part of the content bundle it is not as important as `additionalWebspaces` which will be used by any entity using content bundle with routing.

### DIFF
--- a/Content/Domain/Model/WebspaceTrait.php
+++ b/Content/Domain/Model/WebspaceTrait.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\ContentBundle\Content\Domain\Model;
+
+trait WebspaceTrait
+{
+    /**
+     * @var string|null
+     */
+    protected $mainWebspace;
+
+    public function getMainWebspace(): ?string
+    {
+        return $this->mainWebspace;
+    }
+
+    public function setMainWebspace(?string $mainWebspace): void
+    {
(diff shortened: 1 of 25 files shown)


Content/Domain/Model/WebspaceTrait.php
```diff
@@ -0,0 +1,63 @@
...
+{
+    /**
+     * @var string|null
+     */
+    protected $mainWebspace;
+
+    /**
+     * @var string[]|null
+     */
+    protected $additionalWebspaces;
```
alexander-schranz, line 26: As discussed with @chirimoya we will go with a `one-to-many` relation.

The content bundle will provide also a `AdditionalWebspaceTrait` content bundle so a article bundle for example has then 3 entities:

```
Article
ArticleDimensionContent
ArticleDimensionContentWebspace
```

The `ArticleDimensionContentWebspace` will have `id`, `dimensionContent` and `webspace` field.

/cc @TheCadien do you want to have a look at this?

TheCadien:
@alexander-schranz i do that

PR #216, opened 2022-01-25
Fix availableLocales after unpublish

alexander-schranz:
The `availableLocacles` should represent the locales which currently are available at this stage. When unpublish the locale need to be removed from it.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,0 +1,22 @@
+parameters:
+	ignoreErrors:
+		-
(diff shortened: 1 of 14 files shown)


phpstan-baseline.neon
```diff
@@ -0,0 +1,22 @@
+parameters:
```
alexander-schranz, line 1: will update phpstan in another issue as this seems to work in phpstan 1.0

wachterjohannes (changes requested):
Linting is failing

wachterjohannes (approved):

PR #217, opened 2022-01-28
Update phpstan to new version

alexander-schranz:


Content/Domain/Model/DimensionContentInterface.php
```diff
@@ -66,14 +66,32 @@ public function isMerged(): bool;
...
+     * }
      */
     public static function getDefaultDimensionAttributes(): array;
 
     /**
-     * @param mixed[] $dimensionAttributes
+     * @internal should only be used by content bundle services
+     *
+     * TODO there is an edge case where locale: null is given and locale: null returned by DimensionContentQueryEnhancer.
+     *      Find a way to set locale return type by
```
alexander-schranz, line 82: https://github.com/phpstan/phpstan/issues/5591#issuecomment-1021297170

Beneath is correctly handled by phpstan but it does add an other bounded error.

PR #218, opened 2022-03-29
Add missing joins to examples list XML

ajdinmore:
Just a little tweak so nobody else sits there incorrectly trying to figure out what's special about case-property that makes cget fail when it's removed from the Example list.  (Plus removal of a redundancy in ExampleDimensionContent that was niggling at me.)


alexander-schranz (approved):

alexander-schranz:
@ajdinmore Thank you!

PR #226, opened 2022-10-03
Add indexes to improve performance of some queries

alexander-schranz:
While having a look at our projects recipe bundle. I got hit by very slow queries.

Before: ~2023 ms
After: ~20ms (without filters, with tag or category filters ~1300ms see #227 for more improvements)

### DIFF
--- a/Content/Infrastructure/Sulu/SmartContent/Repository/ContentDataProviderRepository.php
+++ b/Content/Infrastructure/Sulu/SmartContent/Repository/ContentDataProviderRepository.php
@@ -462,7 +462,6 @@ class ContentDataProviderRepository implements DataProviderRepositoryInterface
 
         return $this->entityManager->createQueryBuilder()
             ->select(self::CONTENT_RICH_ENTITY_ALIAS . '.' . $this->getEntityIdentifierFieldName() . ' as id')
-            ->distinct()
             ->from($this->contentRichEntityClass, self::CONTENT_RICH_ENTITY_ALIAS)
             ->innerJoin(self::CONTENT_RICH_ENTITY_ALIAS . '.dimensionContents', self::LOCALIZED_DIMENSION_CONTENT_ALIAS)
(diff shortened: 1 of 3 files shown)


Content/Infrastructure/Sulu/SmartContent/Repository/ContentDataProviderRepository.php
```diff
@@ -460,7 +459,6 @@ protected function createEntityIdsQueryBuilder(string $locale): QueryBuilder
 
         return $this->entityManager->createQueryBuilder()
             ->select(self::CONTENT_RICH_ENTITY_ALIAS . '.' . $this->getEntityIdentifierFieldName() . ' as id')
-            ->distinct()
```
alexander-schranz, line 465: This hurts the performance a lot but is not necessary. All joins of filters should be created the way that no distinct is required here else we will run into a lot of performance issues.

Content/Infrastructure/Sulu/SmartContent/Repository/ContentDataProviderRepository.php
```diff
@@ -402,6 +401,8 @@ protected function setSortByJoins(QueryBuilder $queryBuilder): array
      */
     protected function appendRelation(QueryBuilder $queryBuilder, string $relation, array $values, string $operator, string $alias): array
     {
+        $queryBuilder->distinct(); // TODO remove distinct and replace joins with subselect filter see: https://github.com/sulu/SuluContentBundle/pull/226
```
alexander-schranz, line 406: created an issue for this. here some improvements could be made: https://github.com/sulu/SuluContentBundle/issues/227

Content/Infrastructure/Sulu/SmartContent/Repository/ContentDataProviderRepository.php
alexander-schranz: We should in future have a look at the queries that we not longer have todo `distinct` as it seems have bad effects on this kind of queries. See #227

PR #238, opened 2023-03-21
Add Shadow Locale to Settings Tab

alexander-schranz:
To provide a shadow locale feature we add a new ShadowInterface for this which provides this feature. Then new `shadowLocale` and `shadowLocales` feature is added to the DimensionContent entities and added to the Settings tab.

In upcoming Pull Request the shadow behaviour while publishing should be implemented and checked how we handle the `url` here.


Content/Domain/Model/ShadowTrait.php
```diff
@@ -0,0 +1,68 @@
...
+{
+    /**
+     * @var string|null
+     */
+    protected $shadowLocale;
+
+    /**
+     * @var string[]|null
+     */
+    protected $shadowLocales = null;
```
Prokyonn, line 26: Wouldn't it be more convenient to have the default value set to an empty array, then we could omit the check in line 46

alexander-schranz: @Prokyonn there should not exis default values in the dimensions. Not set values are null. This make the dimension table more readable and see which dimension properties are used in which dimension itself and make merging of the dimension easier. So `shadowLocales` will be null on the localized dimension as it should never exist there and only set on the unlocalized dimension.

Same already is done for `availableLocales`.

Content/Application/ContentNormalizer/Normalizer/ShadowNormalizer.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        if (!$object instanceof ShadowInterface) {
+            return $normalizedData;
+        }
+
+        Assert::isInstanceOf($object, DimensionContentInterface::class);
+
+        $normalizedData['shadowOn'] = null !== $object->getShadowLocale();
+        $normalizedData['shadowLocales'] = $normalizedData['shadowLocales'] ?? [];
+        $normalizedData['contentLocales'] = $object->getAvailableLocales() ?? []; // TODO should be changed in Sulu Core (PageSettingsShadowLocaleSelect.js)
```
alexander-schranz, line 32: Issue: https://github.com/sulu/sulu/issues/7040

Resources/translations/admin.de.json
```diff
@@ -5,6 +5,10 @@
     "sulu_content.published": "Veröffentlicht am",
     "sulu_content.author": "Autor",
     "sulu_content.authored_date": "Verfasst am",
+    "sulu_content.shadow_page": "Shadow Seite",
+    "sulu_content.enable_shadow_page": "Shadow Seite aktivieren",
+    "sulu_content.enable_shadow_page_info_text": "",
+    "sulu_content.shadow_locale": "Shadow sprache",
```
Prokyonn:
```suggestion
    "sulu_content.shadow_locale": "Shadow Sprache",
```

Resources/translations/admin.de.json
```diff
@@ -5,6 +5,10 @@
     "sulu_content.published": "Veröffentlicht am",
     "sulu_content.author": "Autor",
     "sulu_content.authored_date": "Verfasst am",
+    "sulu_content.shadow_page": "Shadow Seite",
+    "sulu_content.enable_shadow_page": "Shadow Seite aktivieren",
+    "sulu_content.enable_shadow_page_info_text": "",
```
Prokyonn:
```suggestion
    "sulu_content.enable_shadow_page_info_text": "Diese Funktion ist deaktiviert, wenn die aktuelle Sprache bereits als Shadow verwendet wird.",
```

Prokyonn (approved):

PR #239, opened 2023-03-23
Fix showing of ghost locale identicator for Example

alexander-schranz:
The `ghostLocale` identicator is currently show always for the ExampleTestBundle but should only be shown when it was not yet translated into that language. This adopt the query for not give back a `ghostLocale` if current `locale` is available.

### DIFF
--- a/Tests/Application/ExampleTestBundle/Resources/config/lists/examples.xml
+++ b/Tests/Application/ExampleTestBundle/Resources/config/lists/examples.xml
@@ -25,7 +25,7 @@
             <entity-name>ghostDimensionContent</entity-name>
             <field-name>Sulu\Bundle\ContentBundle\Tests\Application\ExampleTestBundle\Entity\Example.dimensionContents</field-name>
             <method>LEFT</method>
-            <condition>ghostDimensionContent.locale = unlocalizedDimensionContent.ghostLocale AND ghostDimensionContent.stage = 'draft'</condition>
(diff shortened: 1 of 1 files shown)


Tests/Application/ExampleTestBundle/Resources/config/lists/examples.xml
```diff
@@ -25,7 +25,7 @@
             <entity-name>ghostDimensionContent</entity-name>
             <field-name>Sulu\Bundle\ContentBundle\Tests\Application\ExampleTestBundle\Entity\Example.dimensionContents</field-name>
             <method>LEFT</method>
-            <condition>ghostDimensionContent.locale = unlocalizedDimensionContent.ghostLocale AND ghostDimensionContent.stage = 'draft'</condition>
+            <condition>dimensionContent.locale IS NULL AND ghostDimensionContent.locale = unlocalizedDimensionContent.ghostLocale AND ghostDimensionContent.stage = 'draft'</condition>
```
alexander-schranz, line 28: the `ref=` here would have a requiremend to `unlocalizedDimensionContent` and to `dimensionContent` join but we are currently not supporting soemthing like `ref="dimensionContent, unlocalizedDimensionContent"` or is there other way to reference other joins in here? Maybe a `<join-ref name="" />` instead of `ref` attribute would make sense here?

wachterjohannes: hmm not possible currently like the idea of an own node but would put that into the joins node:

```
<joins>
  <joins ref="...."/>
  <joins ref="...."/>
  <join ...></join>
</joins>
```

wachterjohannes (approved):

PR #249, opened 2023-05-31
Enable preview for shadow entities

Prokyonn:
This PR removes the preview condition for shadowlocales and adds a temporary solution for the preview.

This solution is not really good, as we need to load the dimensionContent twice for shadowLocales, but it should be sufficient for now.


Content/Infrastructure/Sulu/Preview/ContentObjectProvider.php
```diff
@@ -193,6 +194,17 @@ protected function resolveContent(ContentRichEntityInterface $contentRichEntity,
                 ]
             );
 
+            // unfortunately we can only check if it is a shadow after the dimensionContent was loaded
+            if ($resolvedDimensionContent instanceof ShadowInterface && $resolvedDimensionContent->getShadowLocale()) {
+                $resolvedDimensionContent = $this->contentResolver->resolve(
```
alexander-schranz: Shoukd we do here return $this->resolvecContent. So it is recursicely resolved also when the target shadow is itself a shadow.

alexander-schranz (approved):

PR #254, opened 2023-08-29
Do not process SettingsFormPass without AdminBundle

maikrosenthal:
In our setup, the AdminBundle is only loaded in the admin context. Therefore, the following error is thrown in the frontend: "You have requested a non-existent parameter "sulu_admin.forms.directories"."

### DIFF
--- a/DependencyInjection/Compiler/SettingsFormPass.php
+++ b/DependencyInjection/Compiler/SettingsFormPass.php
@@ -22,6 +22,10 @@ class SettingsFormPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container): void
     {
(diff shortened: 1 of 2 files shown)


alexander-schranz:
Hello @maikrosenthal 👋 ,

The SuluAdminBundle is enabled in both context since Sulu 2.3, if your application did not yet update to this we recommend it: https://github.com/sulu/sulu/blob/2.3.0/config/bundles.php. It is not longer recommended to build on top of sulu.context flag any logic and services. The sulu.context will be removed when there is no longer PHPCR and replaced by the content bundle.

To make this more clear we can throw here an exception that the `SuluAdminBundle` was not correctly registered.

PR #255, opened 2023-12-21
Add update date to settings tab

martinlagler:


Content/Domain/Model/AuthorInterface.php
```diff
@@ -17,11 +17,17 @@
...
+    public function getLastModified(): ?\DateTime;
+
+    public function setLastModified(?\DateTime $lastModified): void;
+
     public function getAuthor(): ?ContactInterface;
 
     public function setAuthor(?ContactInterface $author): void;
 
-    public function getAuthored(): ?\DateTimeImmutable;
+    public function getAuthored(): ?\DateTime;
```
alexander-schranz: we should always use `DateTimeImmutable`

composer.json
```diff
@@ -14,13 +14,13 @@
         "php": "^7.2 || ^8.0",
         "doctrine/inflector": "^1.4.1 || ^2.0.1",
         "doctrine/collections": "^1.0",
-        "doctrine/orm": "^2.5.3",
+        "doctrine/orm": "^2.5.3 || ^2.8",
```
alexander-schranz: unnecessary change

composer.json
```diff
@@ -14,13 +14,13 @@
...
         "doctrine/persistence": "^1.3 || ^2.0",
         "doctrine/doctrine-bundle": "^1.10 || ^2.0",
         "friendsofsymfony/rest-bundle": "^2.6 || ^3.0",
         "massive/search-bundle": "^2.4",
         "ramsey/uuid": "^3.8 || ^4.0",
-        "sulu/sulu": "^2.4 || ^2.6@dev",
+        "sulu/sulu": "2.6.x-dev",
```
alexander-schranz: we should avoid eventually required upgrade may we can bridge it

composer.json
```diff
@@ -54,7 +54,7 @@
         "phpstan/phpstan-webmozart-assert": "^1.0",
         "phpunit/phpunit": "^8.5.33 || ^9.6.5",
         "qossmic/deptrac-shim": "^0.11.1 || ^0.23.0 || ^1.0",
-        "sulu/automation-bundle": "^2.0@dev",
+        "sulu/automation-bundle": "^2.0@dev || ^2.1",
```
alexander-schranz:
```suggestion
        "sulu/automation-bundle": "^2.0 || ^2.1@dev",
```

composer.json
```diff
@@ -66,7 +66,8 @@
     },
     "conflict": {
         "coduo/php-matcher": "6.0.12",
-        "doctrine/persistence": "1.3.2"
+        "doctrine/persistence": "1.3.2",
+        "doctrine/orm": "2.10.0 || 2.14.2 || >=2.16.0"
```
alexander-schranz: Should not longer be required

```suggestion
        "doctrine/persistence": "1.3.2"
```

alexander-schranz (changes requested):

PR #258, opened 2024-05-28
Add Symfony 7 support

TheCadien:
Allow Symfony 7 Packages.

part of https://github.com/sulu/sulu/issues/7413


composer.json
```diff
@@ -11,7 +11,7 @@
     "homepage": "https://github.com/sulu/SuluContentBundle",
     "license": "MIT",
     "require": {
-        "php": "^7.2 || ^8.0",
+        "php": "^8.2 || ^8.3",
```
alexander-schranz: The CI need to match sulu versions from PHP `8.0` on lowest

alexander-schranz:
Symfony 7 is not run yet. The task bundle blocks it:

```
php-task/task-bundle                             3.0.0     requires  symfony/dependency-injection (^5.4 || ^6.0)
```

@TheCadien you want to tackle the php-task component and bundle pull requests?

Content/Infrastructure/Doctrine/MetadataLoader.php
```diff
@@ -35,15 +35,8 @@
 /**
  * @internal
  */
-final class MetadataLoader implements EventSubscriber
```
alexander-schranz, line 38: Same as for https://github.com/sulu/sulu/pull/7374

alexander-schranz:
The lowest test seems be a problem that we may need register the `MetadataLoader` depending on the `doctrine.orm.listeners.resolve_target_entity` but Subscriber is registered before Listeners. So we may need create a bridge.

In Sulu core we resolved it differently not sure if it was a good idea: https://github.com/sulu/sulu/commit/dacce4ec577ca8a20571c984191385af70340954

alexander-schranz:
@Prokyonn FYI

PR #260, opened 2024-07-16
Decorate structure resolver to support AuthorInterface

Prokyonn:
Sulu core uses the [StructureResolver](https://github.com/sulu/sulu/blob/2.6/src/Sulu/Bundle/WebsiteBundle/Resolver/StructureResolver.php#L115-L118) to set `author` and `authored`

This is not working for the contentbundle, because the DimensionContent entity does not use the `LocalizedAuthorBehavior` interface, but instead the `AuthorInterface` of the content bundle directly.

This MR decorates the original resolver to also support the contentBundle dimensions.

### DIFF
--- a/Content/Infrastructure/Sulu/Structure/DecoratedStructureResolver.php
+++ b/Content/Infrastructure/Sulu/Structure/DecoratedStructureResolver.php
@@ -17,6 +17,9 @@ use Sulu\Bundle\ContentBundle\Content\Domain\Model\AuthorInterface;
 use Sulu\Bundle\WebsiteBundle\Resolver\StructureResolverInterface;
 use Sulu\Component\Content\Compat\StructureInterface;
 
+/**
+ * @internal
+ */
 class DecoratedStructureResolver implements StructureResolverInterface
 {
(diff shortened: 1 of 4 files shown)


Content/Infrastructure/Sulu/Structure/DecoratedStructureResolver.php
```diff
@@ -0,0 +1,48 @@
...
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\ContentBundle\Content\Infrastructure\Sulu\Structure;
+
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\AuthorInterface;
+use Sulu\Bundle\WebsiteBundle\Resolver\StructureResolverInterface;
+use Sulu\Component\Content\Compat\StructureInterface;
+
+class DecoratedStructureResolver implements StructureResolverInterface
```
alexander-schranz:
```suggestion

/**
 * @internal
 */
class DecoratedStructureResolver implements StructureResolverInterface
```

alexander-schranz (changes requested):

PR #262, opened 2024-07-30
Add own Website ContentController to render Content

alexander-schranz:
What do you think to already create a controller inside the ContentBundle. This should make upgrading projects easier later as we don't need to adjust the controllers in the .xml files.

Not sure about the name:

   - DefaultController (I personally find it to general)
   - WebsiteController (already in the Website namespace)
   - CmsController (we never used Cms in our code yet)
   - TemplateController (controller should in future also handle JSON still it configured in a Sulu template, so still would be fine for me)
   - ContentController


Content/UserInterface/Controller/Website/ContentController.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Sulu\Bundle\ContentBundle\Content\UserInterface\Controller\Website;
+
+use Sulu\Bundle\WebsiteBundle\Controller\DefaultController;
+
+/**
+ * TODO this controller will later replace the DefaultController of the WebsiteBundle
+ *      and should be base for all content controllers.
+ */
+class ContentController extends DefaultController
```
alexander-schranz, line 22: Not sure about the name:

 - DefaultController (I personally find it to general)
 - WebsiteController (already in the Website namespace)
 - CmsController (we never used Cms in our code yet)
 - TemplateController (controller should in future also handle JSON still it configured in a Sulu template, so still would be fine for me)
 - ContentController

Prokyonn: `ContentController` sounds good for me

Prokyonn (approved):

PR #268, opened 2024-08-09
Add interface for ContentRichEntityRepository

alexander-schranz:
We require a minimal interface which all content rich entity need to implemented. This interface should then be used to provide basic implementations of:

 - [ ] ContentRouteDefaultsProvider
 - [ ] TeaserProvider
 - [ ] LinkProvider
 - [ ] SmartContentProvider
 - [ ] Selection
 - [ ] Single Selection

It should also replace the whole `DimensionContentRepository` which we sadly use at a lot of cases.


Content/Domain/Repository/ContentRichEntityRepositoryInterface.php
```diff
@@ -0,0 +1,76 @@
...
+
+namespace Sulu\Bundle\ContentBundle\Content\Domain\Repository;
+
+use Sulu\Bundle\ContentBundle\Content\Domain\Model\ContentRichEntityInterface;
+
+/**
+ * @template T of ContentRichEntityInterface
+ *
+ * @phpstan-type ContentRichEntityRepositoryFilters array{
+ *      ids: array<string>,
```
alexander-schranz: this may is a `ids` or `uuids` so not sure about how we handle this. Should we go with `identifiers` @Prokyonn

Content/Domain/Repository/ContentRichEntityRepositoryInterface.php
```diff
@@ -0,0 +1,76 @@
...
+ *      locale?: string|null,
+ *      stage?: string|null,
+ *      categoryIds?: int[],
+ *      categoryKeys?: string[],
+ *      categoryOperator?: 'AND'|'OR',
+ *      tagIds?: int[],
+ *      tagNames?: string[],
+ *      tagOperator?: 'AND'|'OR',
+ *      templateKeys?: string[],
+ *      loadGhost?: bool,
```
alexander-schranz: `ids` / `identifiers` is the only when which need be implemented the other filters are provided via the QueryEnhancer:

```php
            $this->dimensionContentQueryEnhancer->addFilters(
                $queryBuilder,
                'example',
                ExampleDimensionContent::class,
                $filters,
                $sortBys
            );
```

alexander-schranz (commented):
@Prokyonn what do you think about the interface?

Content/Domain/Repository/ContentRichEntityRepositoryInterface.php
alexander-schranz: The ContentRichEntityInterface has a `getId()`: https://github.com/sulu/SuluContentBundle/blob/f36e36d52f01c4da54d80b9b08a64e953e86e91d/Content/Domain/Model/ContentRichEntityInterface.php#L26 but now that we said may Identifier make more sense in case of id vs uuid. But another option is as we already force `getId` in the interface keep `ids` and may also just rename `Artice:uuid` to `Article::id` even if its a uuid. /cc @chirimoya

Prokyonn: Technically speaking, `id` is just the abbreviation of `identifier` 😅 
also writing `['identifier' => xx]` feels more cumbersome than `'id' => xx` 🤔 

Although we may use `uuid`, I think I would still go with `id` as a string.

Prokyonn (approved):

